PR #29, opened 2020-06-02
Added ability to bind epg to static fex port

manofcolombia:
**SUMMARY**
Added ability to configure static port bindings on a fex downstream of a leaf. Fex ports required a different static path.

**ISSUE TYPE**
Feature Pull Request

**ADDITIONAL INFORMATION**
Prior to this addition, the module is unable to configure fex ports because portpath was hard coded to a path specific to leaf ports only. Introducing new, optional fex fields and conditional now allows for fex ports to be configured if desired.

Tested with MSO: 3.0(1i) and ACI: 4.2(4i)

### DIFF
--- a/plugins/modules/mso_schema_site_anp_epg_staticport.py
+++ b/plugins/modules/mso_schema_site_anp_epg_staticport.py
@@ -58,11 +58,6 @@ options:
     description:
     - The leaf of the static port.
     type: str
-  fex_port:
-    description:
-    - Indicate if the static port resides on a fex.
-    type: bool
-    default: no
   fex:
     description:
     - The fex id of the static port.
@@ -140,7 +135,6 @@ EXAMPLES = r'''
     type: port
     pod: pod-1
     leaf: 101
-    fex_port: yes
     fex: 151
     path: eth1/1
     vlan: 126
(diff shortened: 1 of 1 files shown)


plugins/modules/mso_schema_site_anp_epg_staticport.py
```diff
@@ -185,6 +215,8 @@ def main():
         type=dict(type='str', default='port', choices=['port']),
         pod=dict(type='str'),  # This parameter is not required for querying all objects
         leaf=dict(type='str'),  # This parameter is not required for querying all objects
+        fex_port=dict(type=bool, default=False),   # This parameter is not required for querying all objects
```
lhercot, line 218: Change to type='bool' so that it passes sanity testing.

lhercot: Do we need the fex_port attribute?
We should be able to use the fex attribute for that condition.

manofcolombia: I believe we could get away with just using the fex attribute if it wasn't being used to specify that fex attribute is required.
`['fex_port', 'true', ['type', 'pod', 'leaf', 'fex', 'path', 'vlan']]`

If we switched to just the fex attribute, we'd have to find a way to require fex if the value of fex was not none/null/empty. I tried a couple of different ways to do that, but it seems like `required_if` is always looking for a key, value and will not take anything other than a raw string as the value. 
Ex. `['fex', is not none, ['type', 'pod', 'leaf', 'fex', 'path', 'vlan']]` This produces a syntax error

lhercot: But you do not need to do that check anymore because all the other parameters are already required for absent or present right?
If 'fex' exists, we build the path from line 255.

manofcolombia: I see what you are saying. I had not thought about it like that, but you are correct and I was able to completely remove fex_port param.

PR #30, opened 2020-06-03
fix issue 'mso_schema_template_contract_filter does not work for vrf scope #17'

### LINKED ISSUE #17: mso_schema_template_contract_filter does not work for vrf scope
The valid options for contract_scope are:

contract_scope=dict(type='str', choices=['application-profile', 'global', 'tenant', 'vrf'])

MSO permits all of these except 'vrf'. If the user specifies 'vrf', MSO expects the string 'context' to be sent instead.

xinyuezhao:
add test file for module mso_schema_template_contract_filter
fix issue 'mso_schema_template_contract_filter does not work for vrf scope #17'
add function to transform filterRef format

Resolves #17

### DIFF
--- a/plugins/module_utils/mso.py
+++ b/plugins/module_utils/mso.py
@@ -458,6 +458,26 @@ class MSOModule(object):
             'templateName': vrf_dict.group(2),
         }
 
+    def dict_from_ref(self, data):
+        ref_regex = re.compile(r'\/schemas\/(.*)\/templates\/(.*)\/(.*)\/(.*)')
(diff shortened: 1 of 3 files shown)


andbyrne:
The contract_scope fix will only work for new contracts. If someone is updating and specifies 'vrf' for contract_scope, it will still fail. Alternative is to update the module as follows:

    contract_scope = 'context' if module.params['contract_scope'] == 'vrf' else module.params['contract_scope']

PR #53, opened 2020-06-13
Enhancement of the mso_schema_site_anp_epg_domain module and its updated test file.

shrsr:
Enhancements:
1. The domain module creates ANP and EPG  at site level, if they already exist at template level
2. Fail messages have been modified to display appropriate information on its cause
3. Object of domain association and their attributes are consistent


plugins/modules/mso_schema_site_anp_epg_domain.py
```diff
@@ -141,8 +143,7 @@
     resolution_immediacy: pre-provision
     state: present
   delegate_to: localhost
-
```
lhercot, line 144: We need to keep those spaces (and the next 2)

plugins/modules/mso_schema_site_anp_epg_domain.py
```diff
@@ -259,37 +258,99 @@ def main():
     schema_obj = mso.get_obj('schemas', displayName=schema)
     if not schema_obj:
         mso.fail_json(msg="Provided schema '{0}' does not exist".format(schema))
-
     schema_path = 'schemas/{id}'.format(**schema_obj)
     schema_id = schema_obj.get('id')
 
+    # Get template
+    templates = [t.get('name') for t in schema_obj['templates']]
```
lhercot, line 265: We should probably move the schema_obj to .get() like the rest.

tests/integration/targets/mso_schema_site_anp_epg_domain/tasks/main.yml
```diff
@@ -0,0 +1,944 @@
+# Test code for the MSO modules
+# Copyright: (c) 2020, Lionel Hercot (@lhercot) <lhercot@cisco.com>
```
lhercot, line 2: Add a line for your copyright. It is your file after all.

PR #56, opened 2020-06-16
New functionality and modifications for mso_schema_site_anp_epg_staticport

shrsr:
1. Addition of Direct Port Channel option and its attribute Micro Segment VLAN
2. Addition of default values for deployment immediacy and mode 
3. Updated Test File

### DIFF
--- a/plugins/modules/mso_schema_site_anp_epg_staticport.py
+++ b/plugins/modules/mso_schema_site_anp_epg_staticport.py
@@ -47,6 +47,9 @@ options:
   type:
     description:
     - The path type of the static port
+    - vpc is used for a Virtual Port Channel
+    - dpc is used for a Direct Port Channel
+    - port is used for a single interface
     type: str
     choices: [ port, vpc, dpc ]
     default: port
--- a/plugins/modules/mso_schema_site_anp_epg_staticport.py
+++ b/plugins/modules/mso_schema_site_anp_epg_staticport.py
@@ -87,9 +87,9 @@ options:
     type: str
     choices: [ native, regular, untagged ]
     default: untagged
-  micro_segment_vlan:
+  primary_micro_segment_vlan:
     description:
-    - Primary micro-seg VLAN of dpc
+    - Primary micro-seg VLAN of static port.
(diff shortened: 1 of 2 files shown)


plugins/modules/mso_schema_site_anp_epg_staticport.py
```diff
@@ -85,6 +86,11 @@
     - C(untagged) means B(Access (untagged)).
     type: str
     choices: [ native, regular, untagged ]
+    default: untagged
+  micro_segment_vlan:
```
lhercot, line 90: Would use the same as UI like primary_micro_seg_vlan

plugins/modules/mso_schema_site_anp_epg_staticport.py
```diff
@@ -85,6 +86,11 @@
...
+    description:
+    - Primary micro-seg VLAN of dpc
```
lhercot, line 92: Primary micro-seg VLAN is for all 3 so we should be clear that it is for Port, Direct Port Channel (dpc) and Virtual Port Channel (vpc).

plugins/modules/mso_schema_site_anp_epg_staticport.py
```diff
@@ -375,6 +385,9 @@ def main():
         type=path_type,
     )
 
+    if path_type == 'dpc':
```
lhercot, line 388: You should just check if the parameter is not None but update it for all cases.

plugins/modules/mso_schema_site_anp_epg_staticport.py
```diff
@@ -248,6 +256,8 @@ def main():
         portpath = 'topology/{0}/paths-{1}/pathep-[{2}]'.format(pod, leaf, path)
     elif path_type == 'vpc':
         portpath = 'topology/{0}/protpaths-{1}/pathep-[{2}]'.format(pod, leaf, path)
+    elif path_type == 'dpc':
```
lhercot, line 259: This condition looks identical to the ports one, should we fuse the two?

PR #59, opened 2020-06-24
build new module 'mso_schema_template_vrf_contract' with test file

### LINKED ISSUE #41: Add support for vzAny contracts
Create new module mso_schema_vrf_contract to add support for vzAny contracts.
Code from mso_schema_external_epg_contract can be use as a base.

xinyuezhao:
build module 'mso_schema_template_vrf_contract' and its test file
add vzAny attribute in module 'mso_schema_template_vrf'

resolve #41

### DIFF
--- a/plugins/modules/mso_schema_template_vrf_contract.py
+++ b/plugins/modules/mso_schema_template_vrf_contract.py
@@ -68,13 +68,12 @@ options:
     choices: [ absent, present, query ]
     default: present
 seealso:
-- module: mso_schema_template_externalepg
-- module: mso_schema_template_contract_filter
+- module: mso_schema_template_vrf
 extends_documentation_fragment: cisco.mso.modules
 '''
 
 EXAMPLES = r'''
-- name: Add a contract to an VRF
+- name: Add a contract to a VRF
   cisco.mso.mso_schema_template_vrf_contract:
     host: mso_host
     username: admin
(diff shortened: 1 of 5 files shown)


plugins/modules/mso_schema_template_vrf_contract.py
```diff
@@ -0,0 +1,265 @@
...
+    if state == 'query':
+        if not contract:
+            provider_contracts = [dict(contractRef=mso.dict_from_ref(c.get('contractRef')),
+                                  relationshipType='provider') for c in schema_obj.get('templates')[template_idx]['vrfs'][vrf_idx]['vzAnyProviderContracts']]
+            consumer_contracts = [dict(contractRef=mso.dict_from_ref(c.get('contractRef')),
+                                  relationshipType='consumer') for c in schema_obj.get('templates')[template_idx]['vrfs'][vrf_idx]['vzAnyConsumerContracts']]
+            mso.existing = provider_contracts + consumer_contracts
+        elif not mso.existing:
+            mso.fail_json(msg="Contract '{0}' not found".format(contract.get('name')))
+
```
lhercot, line 224: Can we add support for query all contract of a VRF if we don't pass the contract name?

plugins/modules/mso_schema_template_vrf_contract.py
```diff
@@ -0,0 +1,265 @@
...
+                schemaId=contract.get('schema_id'),
+            ),
+        )
+
+        mso.sanitize(payload, collate=True)
+
+        if mso.existing:
+            ops.append(dict(op='replace', path=contract_path, value=mso.sent))
+        else:
+            ops.append(dict(op='add', path=contracts_path + '/-', value=mso.sent))
```
lhercot, line 253: You should probably add the /- directly to contracts_path as it is not used anywhere else so you avoid concatenation here again.

lhercot:
Also we should modify the mso_schema_template_vrf test file to add the vzany test.

plugins/modules/mso_schema_template_vrf_contract.py
lhercot: Forget about this comment, the other modules don't do this anyway.

PR #62, opened 2020-06-25
New module: mso_tenant_site 

shrsr:
This module has the ability to:
1. Associate cloud sites with a tenant and set its options
2. Associate non-cloud sites with a tenant

### DIFF
--- a/plugins/modules/mso_tenant_site.py
+++ b/plugins/modules/mso_tenant_site.py
@@ -351,17 +351,13 @@ def main():
 
     sites = [(s.get('siteId')) for s in mso.query_objs('tenants')[tenant_idx]['siteAssociations']]
 
-    if site_id not in sites:
-        pass
-    else:
+    if site_id in sites:
         site_idx = sites.index((site_id))
         mso.existing = mso.query_objs('tenants')[tenant_idx]['siteAssociations'][site_idx]
 
     if state == 'query':
         if len(sites) == 0:
             mso.fail_json(msg="No site associated with tenant Id {0}".format(tenant_id))
-        elif site_id in sites:
-            mso.existing = mso.query_objs('tenants')[tenant_idx]['siteAssociations'][site_idx]
         elif site_id not in sites and site_id is not None:
             mso.fail_json(msg="Site Id {0} not associated with tenant Id {1}".format(site_id, tenant_id))
         elif site_id is None:
--- a/plugins/modules/mso_tenant_site.py
+++ b/plugins/modules/mso_tenant_site.py
@@ -352,7 +352,7 @@ def main():
     sites = [(s.get('siteId')) for s in mso.query_objs('tenants')[tenant_idx]['siteAssociations']]
 
     if site_id not in sites:
-        mso.existing = {}
+        pass
     else:
         site_idx = sites.index((site_id))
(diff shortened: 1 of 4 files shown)


tests/integration/targets/mso_tenant_site/tasks/main.yml
```diff
@@ -0,0 +1,661 @@
...
+
+- name: Dissociate clouds that are associated with ansible_tenant
+  mso_tenant_site:
+    <<: *mso_info
+    tenant: ansible_test
+    site: '{{ item }}'
+    state: absent
+  loop:
+  - '{{ mso_site | default("ansible_test") }}'
+  - '{{ mso_site | default("aws_ansible_test") }}'
```
lhercot, line 43: This will fail if I pass mso_site variable.

lhercot: Change to 'aws_{{ mso_site | default("ansible_test") }}' and 'azure_{{ mso_site | default("ansible_test") }}'

tests/integration/targets/mso_tenant_site/tasks/main.yml
```diff
@@ -0,0 +1,661 @@
...
+- name: Verify ncs_cm
+  assert:
+    that:
+    - ncs_cm is changed
+
+- name: Associate non-cloud site with ansible_test in normal mode
+  mso_tenant_site:
+    <<: *mso_info
+    tenant: ansible_test
+    site: ansible_test
```
lhercot, line 114: Site name does not follow variable define in site. If I change it by passing mso_site variable this and all other references will break. Fix all 'site' attribute to reflect site definition with variable like 'aws_{{ mso_site | default("ansible_test") }}'

plugins/modules/mso_tenant_site.py
```diff
@@ -0,0 +1,394 @@
...
+            if not aws_trusted:
+                if not aws_access_key:
+                    mso.fail_json(msg="aws_access_key is a required field in untrusted mode.")
+                if not secret_key:
+                    mso.fail_json(msg="secret_key is a required field in untrusted mode.")
+            payload['awsAccount'] = [aws_account]
+
+    sites = [(s.get('siteId')) for s in mso.query_objs('tenants')[tenant_idx]['siteAssociations']]
+
+    if site_id not in sites:
```
lhercot, line 354: As discussed previously, we should not assign mso_existing to {} here. Flip this condition to keep only the else part.

plugins/modules/mso_tenant_site.py
```diff
@@ -0,0 +1,394 @@
...
+        mso.existing = {}
+    else:
+        site_idx = sites.index((site_id))
+        mso.existing = mso.query_objs('tenants')[tenant_idx]['siteAssociations'][site_idx]
+
+    if state == 'query':
+        if len(sites) == 0:
+            mso.fail_json(msg="No site associated with tenant Id {0}".format(tenant_id))
+        elif site_id in sites:
```
lhercot, line 363: unnecessary condition

shrsr: If I remove this condition and query a tenant with no sites associated with it, user encounters the following error:
 **None is not in list\n**

It would be neater to have a check and display the appropriate error.

shrsr: What do you suggest?

PR #68, opened 2020-07-02
Fix mso_schema_site_vrf_region_cidr and add VGW attribute to mso_schema_site_vrf_region_cidr_subnet

### LINKED ISSUE #50: [mso_schema_site_vrf_region_cidr_subnet] Add Azure "Used by VGW" support
On an Azure Site, at least one subnet of a CIDR need to be marked for "Used by VGW" to complete a proper deployment.
This request ask to add this support in the mso_schema_site_vrf_region_cidr_subnet module.

### LINKED ISSUE #66: [mso_schema_template_anp_epg] AttributeError: 'NoneType' object has no attribute 'group'
Module Failure when using state [query, present]

AttributeError: 'NoneType' object has no attribute 'group'

[mso_schema_template_anp_epg.error.txt](https://github.com/CiscoDevNet/ansible-mso/files/4845788/mso_schema_template_anp_epg.error.txt)

lhercot:
Fix mso_schema_site_vrf_region_cidr to automatically create VRF and Region if not present at site level
Add support for VGW attribute in mso_schema_site_vrf_region_cidr_subnet
Remove unused regions attribute from mso_schema_template_vrf
Add test for mso_schema_site_vrf_region_cidr module
Add test for mso_schema_site_vrf_region_cidr_subnet module

Resolves #50 
Resolves #66

### DIFF
--- a/plugins/modules/mso_schema_site_vrf_region_cidr.py
+++ b/plugins/modules/mso_schema_site_vrf_region_cidr.py
@@ -2,6 +2,7 @@
 # -*- coding: utf-8 -*-
 
 # Copyright: (c) 2019, Dag Wieers (@dagwieers) <dag@wieers.com>
+# Copyright: (c) 2020, Lionel Hercot (@lhercot) <lhercot@cisco.com>
 # GNU General Public License v3.0+ (see LICENSE or https://www.gnu.org/licenses/gpl-3.0.txt)
 
 from __future__ import absolute_import, division, print_function
@@ -19,6 +20,7 @@ description:
 - Manage site-local VRF region CIDRs in schema template on Cisco ACI Multi-Site.
 author:
 - Dag Wieers (@dagwieers)
+- Lionel Hercot (@lhercot)
 version_added: '2.8'
 options:
   schema:
@@ -73,7 +75,7 @@ extends_documentation_fragment: cisco.mso.modules
 
 EXAMPLES = r'''
 - name: Add a new site VRF region CIDR
-  mso_schema_template_vrf_region_cidr:
+  cisco.mso.mso_schema_site_vrf_region_cidr:
     host: mso_host
     username: admin
     password: SomeSecretPassword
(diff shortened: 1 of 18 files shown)


plugins/modules/mso_schema_site_vrf_region_cidr.py
```diff
@@ -178,48 +181,93 @@ def main():
...
+                    cidr_idx = cidrs.index(cidr)
+                    # FIXME: Changes based on index are DANGEROUS
+                    cidr_path = '/sites/{0}/vrfs/{1}/regions/{2}/cidrs/{3}'.format(site_template, vrf, region, cidr_idx)
+                    mso.existing = schema_obj.get('sites')[site_idx]['vrfs'][vrf_idx]['regions'][region_idx]['cidrs'][cidr_idx]
+                op_path = '/sites/{0}/vrfs/{1}/regions/{2}/cidrs/-'.format(site_template, vrf, region)
+                payload = new_cidr
 
     if state == 'query':
-        if cidr is None:
+        if region not in regions:
```
xinyuezhao, line 263: what's error msg if vrf doesn't exist at site level but exist at template level?

lhercot: No error message, the VRF will be automatically created at VRF level see code from line 223 to 235.

xinyuezhao: In my understanding, we do create region automatically while it doesn't exist in site level. But when we do query, we give an error msg when there is no site-level region. 
should we do the same in vrf?

tests/integration/targets/mso_schema_template_vrf_contract/tasks/main.yml
```diff
@@ -41,6 +41,15 @@
 #     - https://{{ apic_hostname }}
 #     state: present
```
shrsr, line 43: The Remove Schemas block has been shifted. What's the reason?

lhercot: As discussed this is required to make it work for the new cloud modules.

PR #70, opened 2020-07-03
module for creating selectors/subnets at site level

shrsr:
1. Subnets are defined as selectors in the module

### DIFF
--- a/plugins/modules/mso_schema_site_external_epg_selector.py
+++ b/plugins/modules/mso_schema_site_external_epg_selector.py
@@ -1,0 +1,251 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# Copyright: (c) 2020, Shreyas Srish (@shrsr) <ssrish@cisco.com>
+# GNU General Public License v3.0+ (see LICENSE or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+ANSIBLE_METADATA = {'metadata_version': '1.1',
+                    'status': ['preview'],
+                    'supported_by': 'community'}
+
+DOCUMENTATION = r'''
+---
+module: mso_schema_site_external_epg_selector
+short_description: Manage External EPG selector in schema of cloud sites
+description:
+- Manage External EPG selector in schema of cloud sites on Cisco ACI Multi-Site.
+author:
+- Shreyas Srish (@shrsr)
+options:
+  schema:
+    description:
+    - The name of the schema.
+    type: str
+    required: yes
+  template:
+    description:
+    - The name of the template to change.
+    type: str
+    required: yes
+  external_epg:
+    description:
+    - The name of the External EPG to be managed.
+    type: str
+    required: yes
+  site:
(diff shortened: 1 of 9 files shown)


plugins/modules/mso_schema_site_external_epg_selector.py
```diff
@@ -0,0 +1,251 @@
...
+                templateName=template,
+                externalEpgName=external_epg,
+            )
+        )
+        payload.update(l3outDn='')
+    else:
+        external_epg_idx = external_epgs.index(ext_epg_ref)
+
+    # Get Selector
+    if not payload:
```
lhercot, line 198: I don't think this condition is necessary because if you remove this line you are still in the previous else statement which is equivalent to "if not payload"

plugins/modules/mso_schema_site_external_epg_selector.py
```diff
@@ -0,0 +1,251 @@
...
+    if ext_epg_ref not in external_epgs:
+        op_path = '/sites/{0}/externalEpgs/-'.format(site_template)
+        payload.update(
+            externalEpgRef=dict(
+                schemaId=schema_id,
+                templateName=template,
+                externalEpgName=external_epg,
+            )
+        )
+        payload.update(l3outDn='')
```
lhercot, line 193: If there is no condition for this l3outDn='' then you want to put it in the previous dict assignment.

tests/integration/targets/mso_schema_template_external_epg_selector/tasks/main.yml
```diff
@@ -0,0 +1,454 @@
+# Test code for the MSO modules
+# Copyright: (c) 2020, Lionel Hercot (@lhercot) <lhercot@cisco.com>
```
lhercot, line 2: Don't forget your copyright.

tests/integration/targets/mso_schema_template_external_epg_selector/tasks/main.yml
```diff
@@ -0,0 +1,454 @@
...
+  register: nm_non_existing_template
+
+- name: Verify non_existing_template
+  assert:
+    that:
+    - cm_non_existing_template is not changed
+    - nm_non_existing_template is not changed
+    - cm_non_existing_template == nm_non_existing_template
+    - cm_non_existing_template.msg == nm_non_existing_template.msg == "Provided template 'non-existing-template' does not exist. Existing templates{{':'}} Template 1, Template 2"
+
```
lhercot, line 423: We should have a case for existing template but not associated to the site.

tests/integration/targets/mso_schema_template_external_epg_selector/tasks/main.yml
```diff
@@ -0,0 +1,454 @@
...
+  mso_tenant: 
+    <<: *mso_info
+    tenant: ansible_test
+    users:
+    - '{{ mso_username }}'
+    # sites:
+    # - '{{ mso_site | default("ansible_test") }}'
+    state: present
+
+- name: Remove schemas
```
lhercot, line 36: Move this block before line 26.

PR #75, opened 2020-07-09
Updated Test File for mso_user

shrsr:
1. Added test case for read only user
2. Bumped up test coverage

### DIFF
--- a/plugins/modules/mso_user.py
+++ b/plugins/modules/mso_user.py
@@ -248,14 +248,17 @@ def main():
         if mso.existing:
             if not issubset(mso.sent, mso.existing):
                 # NOTE: Since MSO always returns '******' as password, we need to assume a change
-                if 'password' in mso.proposed:
-                    mso.module.warn("A password change is assumed, as the MSO REST API does not return passwords we do not know.")
-                    mso.result['changed'] = True
+                mso.stdout = 'reached pwd' + '\n'
+                mso.module.warn("A password change is assumed, as the MSO REST API does not return passwords we do not know.")
+                mso.result['changed'] = True
 
                 if module.check_mode:
+                    mso.stdout = 'reached check' + '\n'
                     mso.existing = mso.proposed
                 else:
+                    mso.stdout = 'reached put' + '\n'
                     mso.existing = mso.request(path, method='PUT', data=mso.sent)
+
         else:
             if module.check_mode:
                 mso.existing = mso.proposed
--- a/tests/integration/targets/mso_user/tasks/main.yml
+++ b/tests/integration/targets/mso_user/tasks/main.yml
@@ -28,6 +28,12 @@
     user: ansible_test2
   register: cm_remove_user
 
+- name: Remove user ansible_test_read
+  mso_user:
+    <<: *user_absent
+    user: ansible_test_read
+  register: cm_remove_user2
+
 
 # ADD USER
 - name: Add user (check_mode)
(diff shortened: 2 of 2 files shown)


plugins/modules/mso_user.py
```diff
@@ -248,14 +248,17 @@ def main():
         if mso.existing:
             if not issubset(mso.sent, mso.existing):
                 # NOTE: Since MSO always returns '******' as password, we need to assume a change
-                if 'password' in mso.proposed:
-                    mso.module.warn("A password change is assumed, as the MSO REST API does not return passwords we do not know.")
-                    mso.result['changed'] = True
+                mso.stdout = 'reached pwd' + '\n'
```
lhercot, line 251: Remove all the stdout from the code.

plugins/modules/mso_user.py
```diff
@@ -248,14 +248,17 @@ def main():
         if mso.existing:
             if not issubset(mso.sent, mso.existing):
                 # NOTE: Since MSO always returns '******' as password, we need to assume a change
-                if 'password' in mso.proposed:
```
lhercot, line 251: Why removing this condition?

plugins/modules/mso_user.py
```diff
@@ -248,14 +248,17 @@ def main():
...
-                    mso.module.warn("A password change is assumed, as the MSO REST API does not return passwords we do not know.")
-                    mso.result['changed'] = True
+                mso.stdout = 'reached pwd' + '\n'
+                mso.module.warn("A password change is assumed, as the MSO REST API does not return passwords we do not know.")
+                mso.result['changed'] = True
 
                 if module.check_mode:
```
lhercot, line 255: This condition should probably be adjusted to make sure it only fires when needed.

tests/integration/targets/mso_user/tasks/main.yml
```diff
@@ -116,6 +124,85 @@
...
+  register: cm_add_user2_again
+
+- name: Add user in read only again(check mode)
+  mso_user: 
+    <<: *user_present2
+    user_password: '#123455#123455Ee'
+  check_mode: yes
+  register: cm_add_user2_again
+
+- name: Add user in read only (normal mode)
```
lhercot, line 186: This section should be at line 165 before the again.

tests/integration/targets/mso_user/tasks/main.yml
```diff
@@ -116,6 +124,85 @@
...
+    - cm_add_user2.current.roles[0].accessType == 'readOnly'
+
+- name: Add user in read only again(check mode)
+  mso_user: 
+    <<: *user_present2
+    user_password: '#123455#123455Aa'
+  check_mode: yes  
+  register: cm_add_user2_again
+
+- name: Add user in read only again(check mode)
```
lhercot, line 173: This block is duplicate from previous one and next one. (3 time the same)

tests/integration/targets/mso_user/tasks/main.yml
```diff
@@ -274,6 +361,16 @@
...
+- name: Query our user
+  mso_user:
+    <<: *user_query
+    user: ansible_test_read
+  register: nm_query_user2
+
+- name: Verify query_user2
+  assert:
+    that:
+    - nm_query_user2 is not changed
```
lhercot, line 373: You should check that it is in read mode here.

PR #76, opened 2020-07-09
add more contract details in output result of module 'mso_schema_template_contract_filter'

### LINKED ISSUE #32: add more information to result of mso_schema_template_contract_filter module
SUMMARY
add more information to mso.previous and mso.existing in mso_schema_template_contract_filter module, for example, contract_scope, contract_filter_type, filter type.

ISSUE TYPE
Enhancement

COMPONENT NAME
mso_schema_template_contract_filter

EXPECTED RESULTS
the result[current] and result[previous] should have more information

xinyuezhao:
- update mso_schema_tempalte_contract_filter module to add more contract details in output result
- update corresponding test file to check those contract details

resolve #32


plugins/modules/mso_schema_template_contract_filter.py
```diff
@@ -312,24 +326,28 @@ def main():
...
             ops.append(dict(op='add', path='/templates/{0}/contracts/-'.format(template), value=payload))
+
         else:
             # Contract exists, but may require an update
             if contract_display_name is not None:
                 ops.append(dict(op='replace', path=contract_path + '/displayName', value=contract_display_name))
             ops.append(dict(op='replace', path=contract_path + '/filterType', value=contract_ftype))
             ops.append(dict(op='replace', path=contract_path + '/scope', value=contract_scope))
 
+        mso.existing['displayName'] = contract_display_name if contract_display_name is not None else contract_obj.get('displayName')
```
lhercot, line 337: I think this would be clearer as a normal condition and not a one liner.

plugins/modules/mso_schema_template_contract_filter.py
```diff
@@ -240,20 +240,33 @@ def main():
...
         filters = [f.get('filterRef') for f in schema_obj.get('templates')[template_idx]['contracts'][contract_idx][filter_key]]
         filter_ref = mso.filter_ref(schema_id=filter_schema_id, template=filter_template, filter=filter_name)
         if filter_ref in filters:
             filter_idx = filters.index(filter_ref)
             filter_path = '/templates/{0}/contracts/{1}/{2}/{3}'.format(template, contract, filter_key, filter_name)
-            mso.existing = schema_obj.get('templates')[template_idx]['contracts'][contract_idx][filter_key][filter_idx]
+            mso.existing = contract_obj.get(filter_key)[filter_idx]
+            mso.existing['filterRef'] = mso.dict_from_ref(mso.existing.get('filterRef'))
+            mso.existing['displayName'] = contract_obj.get('displayName')
+            mso.existing['filterType'] = filter_type
```
lhercot, line 253: Aren't we overriding a filterType with our user-input in the query part?

plugins/modules/mso_schema_template_contract_filter.py
```diff
@@ -240,20 +240,33 @@ def main():
...
 
     if state == 'query':
         if contract_idx is None:
             mso.fail_json(msg="Provided contract '{0}' does not exist. Existing contracts: {1}".format(contract, ', '.join(contracts)))
 
         if filter_name is None:
-            mso.existing = schema_obj.get('templates')[template_idx]['contracts'][contract_idx][filter_key]
+            mso.existing = contract_obj.get(filter_key)
+            for filter in mso.existing:
+                filter['filterRef'] = mso.dict_from_ref(filter.get('filterRef'))
```
lhercot, line 264: We use this whole block twice now, should we move it into a function?

plugins/modules/mso_schema_template_contract_filter.py
xinyuezhao: yes, we do get a filtertype from input. because we can only get contract_filter_type from api.
the filter type is parameter only used to decide the filter path and we can't get it from api.
do you think it's better to remove this from existing or just keep it ?

lhercot: We can just keep it.

plugins/modules/mso_schema_template_contract_filter.py
```diff
@@ -334,11 +337,16 @@ def main():
             ops.append(dict(op='replace', path=contract_path + '/filterType', value=contract_ftype))
             ops.append(dict(op='replace', path=contract_path + '/scope', value=contract_scope))
 
-        mso.existing['displayName'] = contract_display_name if contract_display_name is not None else contract_obj.get('displayName')
+        if contract_display_name:
```
xinyuezhao, line 329: tried to use the same function to replace this part, but got some errors. 
trying to find out whether it's feasible

plugins/modules/mso_schema_template_contract_filter.py
```diff
@@ -334,11 +337,16 @@ def main():
...
+            mso.existing['displayName'] = contract_display_name
+        else:
+            mso.existing['displayName'] = contract_obj.get('displayName')
         mso.existing['filterType'] = filter_type
         mso.existing['contractScope'] = contract_scope
         mso.existing['contractFilterType'] = contract_ftype
```
xinyuezhao, line 336: tried to use the same function to replace this part, but got some errors. 
trying to find out whether it's feasible

PR #79, opened 2020-07-17
new module mso_backup

shrsr:
This module can:
Create, delete and query backups in local mode
 
Test file is included

### DIFF
--- a/plugins/modules/mso_backup.py
+++ b/plugins/modules/mso_backup.py
@@ -1,0 +1,184 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# Copyright: (c) 2018, Dag Wieers (@dagwieers) <dag@wieers.com>
+# Copyright: (c) 2020, Shreyas Srish (@shrsr) <ssrish@cisco.com>
+# GNU General Public License v3.0+ (see LICENSE or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+ANSIBLE_METADATA = {'metadata_version': '1.1',
+                    'status': ['preview'],
+                    'supported_by': 'community'}
+
+DOCUMENTATION = r'''
+---
+module: mso_backup
+short_description: Manage backups
+description:
+- Manage backups on Cisco ACI Multi-Site.
+author:
+- Shreyas Srish (@shrsr)
+version_added: '2.8'
+options:
+  location_type:
+    description:
+    - The location of the backup to be stored
+    type: str
+  backup:
+    description:
+    - The name given to the backup
+    type: str
+    aliases: [ name ]
+  description:
+    description:
+    - Brief information about backup
+    type: str
+  state:
+    description:
+    - Use C(present) or C(absent) for adding or removing.
+    - Use C(query) for listing an object or multiple objects.
+    type: str
+    choices: [ absent, present, query ]
+    default: present
+extends_documentation_fragment: cisco.mso.modules
+'''
+
+EXAMPLES = r'''
+- name: Create a new backup
+  cisco.mso.mso_backup:
+    host: mso_host
+    username: admin
+    password: SomeSecretPassword
+    backup: Backup
+    description: via Ansible
+    location_type: local
+    state: present
+  delegate_to: localhost
+
+- name: Remove a Backup
+  cisco.mso.mso_backup:
+    host: mso_host
+    username: admin
+    password: SomeSecretPassword
+    backup: Backup
+    state: absent
+  delegate_to: localhost
+
+- name: Query a backup
+  cisco.mso.mso_backup:
+    host: mso_host
+    username: admin
+    password: SomeSecretPassword
+    backup: Backup
+    state: query
+  delegate_to: localhost
+  register: query_result
+
+- name: Query all backups
+  cisco.mso.mso_backup:
+    host: mso_host
+    username: admin
+    password: SomeSecretPassword
+    state: query
+  delegate_to: localhost
+  register: query_result
+'''
+
+RETURN = r'''
+'''
+
+from ansible.module_utils.basic import AnsibleModule
+from ansible_collections.cisco.mso.plugins.module_utils.mso import MSOModule, mso_argument_spec, issubset
+
+
+def main():
+    argument_spec = mso_argument_spec()
+    argument_spec.update(
+        location_type=dict(type='str'),
+        description=dict(type='str'),
+        backup=dict(type='str', aliases=['name']),
+        state=dict(type='str', default='present', choices=['absent', 'present', 'query']),
+    )
+
+    module = AnsibleModule(
+        argument_spec=argument_spec,
+        supports_check_mode=True,
+    )
+
+    description = module.params.get('description')
+    name = module.params.get('name')
+    location_type = module.params.get('location_type')
+    state = module.params.get('state')
+    backup = module.params.get('backup')
+
+    mso = MSOModule(module)
+
+    backup_id = None
+    path = 'backups'
+    name = backup
+    counter = 0
+    if backup:
+        mso.existing = mso.query_objs(path)
+        if mso.existing:
+            data = mso.existing
+            for x in data:
+                for k, v in x.items():
+                    if k == 'metadata':
+                        if str(name) in v['name']:
+                            counter += 1
+            if counter >= 1:
+                for i in data:
+                    backup_id = i['id']
+                    for k1, v1 in i.items():
+                        if k1 == 'metadata':
+                            if v1['name'] == name + '_' + backup_id:
+                                path = 'backups/{id}'.format(id=backup_id)
+            else:
+                mso.existing = mso.get_obj(path, name=backup)
+
+    if state == 'query':
+        get_backups = mso.query_objs('backups')
+        mso.existing = []
+        if backup:
+            for i in range(0, len(get_backups)):
+                if (str(backup) + '_' + get_backups[i]['id'] == get_backups[i]['metadata']['name']):
+                    mso.existing.append(get_backups[i])
+        else:
+            mso.existing = mso.query_objs(path)
+        mso.exit_json()
+
+    if state == 'absent':
+        mso.previous = mso.existing
+        if mso.existing:
+            if module.check_mode:
+                mso.existing = {}
+            else:
+                mso.existing = mso.request(path, method='DELETE')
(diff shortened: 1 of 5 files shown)


plugins/modules/mso_backup.py
```diff
@@ -0,0 +1,167 @@
...
+    description = module.params.get('description')
+    location_type = module.params.get('location_type')
+    state = module.params.get('state')
+    backup = module.params.get('backup')
+
+    mso = MSOModule(module)
+
+    backup_id = None
+    path = 'backups'
+    count_backups = 0
```
lhercot, line 116: remove count_backups

plugins/modules/mso_backup.py
```diff
@@ -0,0 +1,167 @@
...
+        if mso.existing:
+            data = mso.existing
+            mso.existing = []
+            for index_of_name in range(0, len(data)):
+                backup_id = data[index_of_name]['id']
+                if (str(backup) + '_' + backup_id == data[index_of_name]['metadata']['name']):
+                    path = 'backups/{id}'.format(id=backup_id)
+                    count_backups += 1
+                    mso.existing.append(data[index_of_name])
+            if count_backups < 1:
```
lhercot, line 128: use len(mso.existing) here.

plugins/modules/mso_backup.py
```diff
@@ -0,0 +1,167 @@
...
+    mso.existing = mso.query_objs('backups')
+    if backup:
+        if mso.existing:
+            data = mso.existing
+            mso.existing = []
+            for index_of_name in range(0, len(data)):
+                backup_id = data[index_of_name]['id']
+                if (str(backup) + '_' + backup_id == data[index_of_name]['metadata']['name']):
+                    path = 'backups/{id}'.format(id=backup_id)
+                    count_backups += 1
```
lhercot, line 126: Remove this line.

plugins/modules/mso_backup.py
```diff
@@ -0,0 +1,167 @@
...
+
+    backup_id = None
+    path = 'backups'
+    count_backups = 0
+    mso.existing = mso.query_objs('backups')
+    if backup:
+        if mso.existing:
+            data = mso.existing
+            mso.existing = []
+            for index_of_name in range(0, len(data)):
```
lhercot, line 122: This looks complicated. Isn't it easier to do:
```
for backup_info in data:
    backup_id = backup_info.get('id')
    if backup_info.get('metadata')['name'] == '{0}_{1}'.format(backup, backup_id):
        path = 'backups/{id}'.format(id=backup_id)
        mso.existing.append(backup_info)
```

plugins/modules/mso_backup.py
```diff
@@ -0,0 +1,167 @@
...
+                backup_id = data[index_of_name]['id']
+                if (str(backup) + '_' + backup_id == data[index_of_name]['metadata']['name']):
+                    path = 'backups/{id}'.format(id=backup_id)
+                    count_backups += 1
+                    mso.existing.append(data[index_of_name])
+            if count_backups < 1:
+                mso.existing = mso.get_obj(path, name=backup)
+
+    if state == 'query':
+        pass
```
lhercot, line 132: why do pass and then exit_json?

plugins/modules/mso_backup.py
```diff
@@ -0,0 +1,167 @@
...
+        mso.exit_json()
+
+    if state == 'absent':
+        mso.previous = mso.existing
+        if mso.existing:
```
lhercot, line 137: What happens if you state is absent but you have more than 1 backup returned? Your condition is probably not strong enough .

plugins/modules/mso_backup.py
```diff
@@ -0,0 +1,167 @@
...
+            if module.check_mode:
+                mso.existing = {}
+            else:
+                mso.existing = mso.request(path, method='DELETE')
+        mso.exit_json()
+
+    path = 'backups'
```
lhercot, line 144: This looks like line 115 and 117 why redo this here? Why also not using get_obj before and just using query_objs?

plugins/modules/mso_user.py
```diff
@@ -228,7 +228,6 @@ def main():
         payload = dict(
```
lhercot, line 228: Why do we have mso_user changes in this PR. It is in the other one right?
I am not reviewing those mso_user changes in this PR. Remove them and make sure they are in the other one.

tests/integration/targets/mso_backup/tasks/main.yml
```diff
@@ -0,0 +1,295 @@
+# Test code for the MSO modules
+# Copyright: (c) 2020, Lionel Hercot (@lhercot) <lhercot@cisco.com>
```
lhercot, line 2: Remove me and Dag from this file. I don't think we still use much of the shared code.

tests/integration/targets/mso_backup/tasks/main.yml
```diff
@@ -0,0 +1,295 @@
...
+    mso_info: &mso_info
+      host: '{{ mso_hostname }}'
+      username: '{{ mso_username }}'
+      password: '{{ mso_password }}'
+      validate_certs: '{{ mso_validate_certs | default(false) }}'
+      use_ssl: '{{ mso_use_ssl | default(true) }}'
+      use_proxy: '{{ mso_use_proxy | default(true) }}'
+      output_level: '{{ mso_output_level | default("info") }}'
+
+- name: Remove schemas
```
lhercot, line 25: Do you need any of this changes for testing the backup?
I think you can probably remove from Line25 to Line97.

tests/integration/targets/mso_backup/tasks/main.yml
```diff
@@ -0,0 +1,295 @@
...
+
+- name: Ensure Template 1 with AP1 exists
+  mso_schema_template_anp:
+    <<: *mso_info 
+    schema: '{{ mso_schema | default("ansible_test") }}'
+    template: Template 1
+    anp: AP1
+    state: present
+
+# Create Backup TO DO MSO Error?
```
lhercot, line 98: TO DO MSO Error?

tests/integration/targets/mso_backup/tasks/main.yml
```diff
@@ -0,0 +1,295 @@
...
+- name: Create Backup1 in check mode
+  mso_backup:
+    <<: *mso_info 
+    backup: Backup1
+    description: via Ansible
+    location_type: local
+    state: present
+  register: b_cm
```
lhercot, line 106: can you use longer variable names? I would use something like:cm_add_backup,  nm_add_backup, 
 cm_add_backup2
Make this change to all the test which don't have readable names.

tests/integration/targets/mso_backup/tasks/main.yml
```diff
@@ -0,0 +1,295 @@
...
+    schema: '{{ mso_schema | default("ansible_test") }}'
+    template: Template 1
+    anp: AP1
+    state: present
+
+# Create Backup TO DO MSO Error?
+- name: Create Backup1 in check mode
+  mso_backup:
+    <<: *mso_info 
+    backup: Backup1
```
lhercot, line 102: Could you prefix this with ansible_ like ansible_backup1?
This will avoid conflict in case of other people using same setup.
Apply this to the other cases too.

tests/integration/targets/mso_backup/tasks/main.yml
```diff
@@ -0,0 +1,295 @@
...
+  mso_backup:
+    <<: *mso_info 
+    state: query
+  ignore_errors: yes
+  register: b_qm_all
+
+- name: Verify b_qm_all
+  assert:
+    that:
+    - b_qm_all is not changed
```
lhercot, line 196: You should check the names of the backups you already know or at least that the number is there.

tests/integration/targets/mso_backup/tasks/main.yml
```diff
@@ -0,0 +1,295 @@
...
+    state: present
+
+- name: Ensure Template 1 with AP1 exists
+  mso_schema_template_anp:
+    <<: *mso_info 
+    schema: '{{ mso_schema | default("ansible_test") }}'
+    template: Template 1
+    anp: AP1
+    state: present
+
```
lhercot, line 97: Shouldn't you have a step to clean previous backups with same name? Or it does not matter to your tests?

tests/integration/targets/mso_backup/tasks/main.yml
```diff
@@ -0,0 +1,295 @@
...
+  ignore_errors: yes
+  
+- name: Remove Backup2
+  mso_backup:
+    <<: *mso_info 
+    backup: Backup2
+    state: absent
+  ignore_errors: yes
+
+# Add more data
```
lhercot, line 267: Why this test?

tests/integration/targets/mso_user/tasks/main.yml
```diff
@@ -19,15 +19,13 @@
     use_ssl: '{{ mso_use_ssl | default(true) }}'
     use_proxy: '{{ mso_use_proxy | default(true) }}'
     output_level: '{{ mso_output_level | default("info") }}'
-    user: ansible_test
+    user: '{{ item }}'
```
lhercot, line 22: Those changes should not be part of this PR.

plugins/modules/mso_backup.py
shrsr: I think mso.get_obj(path, name=backup) is empty because the name attribute in the response body is in metadata and not with id of the backup. Hence, I use query_objs..

"backups": [
    {
      "id": "20200718030540",
      "sizeKB": 790.19,
      "timestamp": "2020-07-18T03:05:40",
      "metadata": {
        "description": "",
        "name": "Backup100_20200718030540"
      }
    }

shrsr: path = 'backups' is like a reset button in this case. We need this because we only use Post method, which requires the main path. If we had Put/Patch we could've avoided it

tests/integration/targets/mso_backup/tasks/main.yml
shrsr: I did add removal steps at the bottom, but yeah I'll move it to the beginning. Will have to see how we can manipulate our next test file related to backups if any.

plugins/modules/mso_backup.py
lhercot: But then why do you use it here. It will clear out the mso.existing which you already have?
If path is always backup for the POST, just put it in the line 161 mso.request you don't need a variable single use.

PR #104, opened 2020-08-24
[mso_schema_site_bd_subnet] fix issue of failing to deploy site bd subnets to APIC

### LINKED ISSUE #93: [mso_schema_site_bd_subnet] Site BD subnet is not being pushed to APIC after deployment
I can use mso_schema_site_bd_subnet module to configure site BD subnet, it would show up in MSO np. However that BD subnet is not being pushed to APIC after schema deployment....If I manually deleted that site BD subnet, added it back thru MSO, then it will show up in APIC after deployment so looks like there is something missing with this mso_schema_site_bd_subnet.

xinyuezhao:
- fix issue of failing to deploy site bd subnets to APIC
- implement test file for mso_schema_site_bd_subnet module

resolve #93

### DIFF
--- a/plugins/modules/mso_schema_site_bd_subnet.py
+++ b/plugins/modules/mso_schema_site_bd_subnet.py
@@ -259,14 +259,6 @@ def main():
         if not mso.existing:
             if description is None:
                 description = subnet
-            if scope is None:
-                scope = 'private'
-            if shared is None:
-                shared = False
-            if no_default_gateway is None:
-                no_default_gateway = False
-            if querier is None:
-                querier = False
 
         payload = dict(
             ip=subnet,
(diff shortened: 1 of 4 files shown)


plugins/modules/mso_schema_site_bd_subnet.py
```diff
@@ -189,6 +190,25 @@ def main():
...
+
+    # Get template BDs
+    template_bds = [b.get('name') for b in schema_obj.get('templates')[template_idx]['bds']]
+
+    # Get template BD
+    if bd not in template_bds:
+        mso.fail_json(msg="Provided BD '{0}' does not exist. Existing template BDs: {1}".format(bd, ', '.join(template_bds)))
+    template_bd_idx = template_bds.index(bd)
+    template_bd = schema_obj.get('templates')[template_idx]['bds'][template_bd_idx]
+    if template_bd.get('l2Stretch') is not False and state == 'present':
```
lhercot, line 207: is not False might be better expressed with is True

PR #106, opened 2020-08-27
fix examples in the  documentation

xinyuezhao:
fix some examples in mso_user module and mso_schema_template_l3out module


lhercot (approved):
LGTM

Issue #89, opened 2020-08-05
New module to support BD migrate feature 


amhunt-at-cisco:
MSO supports migrating a BD between templates with a unique API call, however a new module would need to be written to support that with Ansible. This would be an enhancement - please reach out to amhunt@cisco.com for details. 

Customer_request_label would be amhunt@cisco.com

sacascio:
@lhercot do you know when this feature will be available?

lhercot:
Hi Salvatore,

This enhancement is now in our sprint pipeline so it should be implemented in the next 2-3 weeks.

Kind regards,

[http://www.cisco.com/c/dam/m/en_us/signaturetool/images/banners/standard/01_standard_ciscoblue.png]
Lionel Hercot
Technical Marketing Engineer
Product Development
lhercot@cisco.com<mailto:lhercot@cisco.com>
Tel: +1 408 525 4824
Mobile: +1 408 410 7594
Cisco Systems, Inc.
3625 Cisco Way
San Jose, California 95134
United States
Cisco.com<http://www.cisco.com/>
Download vCard<http://wwwin-tools.cisco.com/dir/vcard/undefined.vcf>
[http://www.cisco.com/assets/swa/img/thinkbeforeyouprint.gif]
Think before you print.
This email may contain confidential and privileged material for the sole use of the intended recipient. Any review, use, distribution or disclosure by others is strictly prohibited. If you are not the intended recipient (or authorized to receive for the recipient), please contact the sender by reply email and delete all copies of this message.
Please click here<http://www.cisco.com/c/en/us/about/legal/terms-sale-software-license-agreement/company-registration-information.html> for Company Registration Information.

PR #122, opened 2020-11-04
[mso_schema_template_migrate ] solves migration of BDs and EPGs across templates by creating new module - Object migration solves issue #89

anvitha-jain:
The issue #89 is solved by creating a new module to support the migration of Bridge Domains (BDs) and End Point Groups (EPGs) between templates in Schema.

### DIFF
--- a/plugins/modules/mso_schema_template_migrate.py
+++ b/plugins/modules/mso_schema_template_migrate.py
@@ -1,0 +1,246 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# Copyright: (c) 2020, Anvitha Jain (@anvitha-jain) <anvjain@cisco.com>
+# GNU General Public License v3.0+ (see LICENSE or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+ANSIBLE_METADATA = {'metadata_version': '1.1',
+                    'status': ['preview'],
+                    'supported_by': 'community'}
+
+DOCUMENTATION = r'''
+---
+module: mso_schema_template_migrate
+short_description: Migrate Bridge Domains (BDs) and EPGs between templates
+description:
+- Migrate BDs and EPGs between templates of same and different schemas.
+author:
+- Anvitha Jain (@anvitha-jain)
+options:
+  schema:
+    description:
+    - The name of the schema.
+    type: str
+    required: yes
+  template:
+    description:
+    - The name of the template.
+    type: str
+    required: yes
+  bds:
+    description:
+    - The name of the BDs to migrate.
+    type: list
+    elements: str
+  epgs:
+    description:
+    - The name of the EPGs and the ANP it is in to migrate.
+    type: list
+    elements: dict
+    suboptions:
+      epg:
+        description:
+        - The name of the EPG to migrate.
+        type: str
+        required: yes
+      anp:
+        description:
+        - The name of the anp to migrate.
+        type: str
+        required: yes
+  target_schema:
+    description:
+    - The name of the target_schema.
+    type: str
+    required: yes
+  target_template:
+    description:
+    - The name of the target_template.
+    type: str
+    required: yes
+  state:
+    description:
+    - Use C(present) for adding.
+    type: str
+    default: present
+extends_documentation_fragment: cisco.mso.modules
+'''
+
+EXAMPLES = r'''
+- name: Migration of objects between templates of same schema
+  mso_schema_template_migrate:
+    host: mso_host
+    username: admin
+    password: SomeSecretPassword
+    schema: Schema 1
+    template: Template 1
+    target_schema: Schema 1
+    target_template: Template 2
+    bds:
+    - BD
+    epgs:
+    - epg: EPG1
+      anp: ANP
+    state: present
+  delegate_to: localhost
+
+- name: Migration of objects between templates of different schema
+  mso_schema_template_migrate:
+    host: mso_host
+    username: admin
+    password: SomeSecretPassword
+    schema: Schema 1
+    template: Template 1
+    target_schema: Schema 2
+    target_template: Template 2
+    bds:
+    - BD
+    epgs:
+    - epg: EPG1
+      anp: ANP
+    state: present
+  delegate_to: localhost
+
+- name: Migration of BD object between templates of same schema
+  mso_schema_template_migrate:
+    host: mso_host
+    username: admin
+    password: SomeSecretPassword
+    schema: Schema 1
+    template: Template 1
+    target_schema: Schema 1
+    target_template: Template 2
+    bds:
+    - BD
+    - BD1
+    state: present
+  delegate_to: localhost
+
+- name: Migration of BD object between templates of different schema
+  mso_schema_template_migrate:
+    host: mso_host
+    username: admin
+    password: SomeSecretPassword
+    schema: Schema 1
+    template: Template 1
+    target_schema: Schema 2
+    target_template: Template 2
+    bds:
+    - BD
+    - BD1
+    state: present
+  delegate_to: localhost
+
+- name: Migration of EPG objects between templates of same schema
+  mso_schema_template_migrate:
+    host: mso_host
+    username: admin
+    password: SomeSecretPassword
+    schema: Schema 1
+    template: Template 1
+    target_schema: Schema 2
+    target_template: Template 2
+    epgs:
+    - epg: EPG1
+      anp: ANP
+    - epg: EPG2
+      anp: ANP2
+    state: present
+  delegate_to: localhost
+
+- name: Migration of EPG objects between templates of different schema
+  mso_schema_template_migrate:
+    host: mso_host
+    username: admin
+    password: SomeSecretPassword
+    schema: Schema 1
+    template: Template 1
+    target_schema: Schema 2
+    target_template: Template 2
+    epgs:
+    - epg: EPG1
+      anp: ANP
+    - epg: EPG2
+      anp: ANP2
+    state: present
+  delegate_to: localhost
+'''
+
+RETURN = r'''
+'''
+
+from ansible.module_utils.basic import AnsibleModule
+from ansible_collections.cisco.mso.plugins.module_utils.mso import MSOModule, mso_argument_spec, mso_object_migrate_spec
+
+
+def main():
+    argument_spec = mso_argument_spec()
+    argument_spec.update(
+        schema=dict(type='str', required=True),
+        template=dict(type='str', required=True),
+        bds=dict(type='list', elements='str'),
+        epgs=dict(type='list', elements='dict', options=mso_object_migrate_spec()),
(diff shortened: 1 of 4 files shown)


tests/integration/targets/mso_schema_template_migrate/tasks/main.yml
```diff
@@ -0,0 +1,221 @@
+# Test code for the MSO modules
+# Copyright: (c) 2020, Anvitha Jain (@anvitha-jain) <anvjain@cisco.com>
+#
```
xinyuezhao, line 3: you may want to remove the empty line

plugins/modules/mso_schema_template_migrate.py
```diff
@@ -0,0 +1,173 @@
...
+from ansible.module_utils.basic import AnsibleModule
+from ansible_collections.cisco.mso.plugins.module_utils.mso import MSOModule, mso_argument_spec, mso_reference_spec, mso_subnet_spec, mso_dhcp_spec
+
+
+def main():
+    argument_spec = mso_argument_spec()
+    argument_spec.update(
+        schema=dict(type='str', required=True),
+        template=dict(type='str', required=True),
+        bds=dict(type='list', elements='str'),  # This parameter is not required for querying all objects
```
xinyuezhao, line 109: do we support query in this module?
if not, you may want to remove comment line here

plugins/modules/mso_schema_template_migrate.py
```diff
@@ -0,0 +1,173 @@
...
+        target_schema=dict(type='str', required=True),
+        target_template=dict(type='str', required=True),
+        state=dict(type='str', default='present'),
+    )
+
+    module = AnsibleModule(
+        argument_spec=argument_spec,
+        supports_check_mode=True,
+        required_if=[
+            ['state', 'absent', ['bds', 'epgs']],
```
xinyuezhao, line 120: if I understand correctly, we only support present.
then maybe we don't need this line.

plugins/modules/mso_schema_template_migrate.py
```diff
@@ -0,0 +1,173 @@
...
+            ['state', 'present', ['bds', 'epgs']],
```
xinyuezhao, line 121: for 'present' state, are 'bds' and 'epgs' both required?
if both of them are required, we can just add 'required=True' in argument_spec.update() since we only have 'present' state.
if they are not required, just remove this line

plugins/modules/mso_schema_template_migrate.py
```diff
@@ -0,0 +1,173 @@
...
+                anps_payload.append(dict(name=anp, epgs=epgs_payload))
+
+            payload = dict(
+                targetSchemaId=target_schema_id,
+                targetTemplateName=target_template,
+                bds=bds_payload,
+                anps=anps_payload,
+            )
+
+            template = template.replace(' ', '%20')
```
xinyuezhao, line 163: did you check in swagger and does it replace the space with "%20"?

tests/integration/targets/mso_schema_template_migrate/tasks/main.yml
```diff
@@ -0,0 +1,221 @@
...
+    host: '{{ mso_hostname }}'
+    username: '{{ mso_username }}'
+    password: '{{ mso_password }}'
+    validate_certs: '{{ mso_validate_certs | default(false) }}'
+    use_ssl: '{{ mso_use_ssl | default(true) }}'
+    use_proxy: '{{ mso_use_proxy | default(true) }}'
+    output_level: '{{ mso_output_level | default("info") }}'
+    schema: '{{ item }}'
+    tenant: ansible_test
+    template: Template_1
```
xinyuezhao, line 94: since you tried to replace space with '%20' to avoid api error, you may want to test for that.
so maybe use "Template 1" instead

tests/integration/targets/mso_schema_template_migrate/tasks/main.yml
```diff
@@ -0,0 +1,221 @@
...
+    <<: *mso_info
+    schema: '{{ mso_schema | default("ansible_test") }}'
+    template: Template_1
+    target_schema: '{{ mso_schema | default("ansible_test_2") }}'
+    target_template: Template_1
+    bds:
+    - ansible_test_1
+    epgs:
+    - epg: ansible_test_1
+      anp: ANP
```
xinyuezhao, line 218: Could it be possible that we only migrate epgs or bds?
if it's possible, can we test it?

tests/integration/targets/mso_schema_template_migrate/tasks/main.yml
```diff
@@ -0,0 +1,221 @@
...
+  register: cm_add_epg
+
+- name: Add EPG 2 (normal mode)
+  mso_schema_template_anp_epg:
+    <<: *epg_present
+    anp: ANP2
+    schema: '{{ mso_schema | default("ansible_test") }}'
+    epg: ansible_test_2
+
+- name: Migration of objects between templates
```
xinyuezhao, line 207: you may want to add another test case to trigger line 147 in mso_schema_template_migrate.py since it's still missing.

tests/integration/targets/mso_schema_template_migrate/tasks/main.yml
shrsr: I second that

plugins/modules/mso_schema_template_migrate.py
```diff
@@ -0,0 +1,173 @@
...
+    type: str
+    required: yes
+  bds:
+    description:
+    - The name of the BDs to migrate.
+    type: list
+    elements: str
+  epgs:
+    description:
+    - The name of the EPGs to migrate.
```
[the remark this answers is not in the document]
shrsr: anp is not used as a parameter is it?

plugins/modules/mso_schema_template_migrate.py
shrsr: I think it would be good to add query.
Once migration is done, user may want to check that the migration has indeed taken place.

plugins/modules/mso_schema_template_migrate.py
shrsr: I second that

shrsr (changes requested):

plugins/modules/mso_schema_template_migrate.py
```diff
@@ -0,0 +1,173 @@
...
+    password: SomeSecretPassword
+    schema: Schema 1
+    template: Template 1
+    target_schema: Schema 1
+    target_template: Template 2
+    bds:
+    - BD
+    epgs:
+    - epg: EPG1
+    anp: ANP
```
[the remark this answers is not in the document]
anvitha-jain: anp is parameter for epg

plugins/modules/mso_schema_template_migrate.py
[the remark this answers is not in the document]
anvitha-jain: anp is EPGs parameter.
name of the EPGS and includes anps also

tests/integration/targets/mso_schema_template_migrate/tasks/main.yml
anvitha-jain: All the other modules are written in this format. @xinyuezhao @shrsr

tests/integration/targets/mso_schema_template_migrate/tasks/main.yml
anvitha-jain: yes, made changes and tested.

plugins/modules/mso_schema_template_migrate.py
anvitha-jain: Done

plugins/modules/mso_schema_template_migrate.py
```diff
@@ -0,0 +1,173 @@
...
+  bds:
+    description:
+    - The name of the BDs to migrate.
+    type: list
+    elements: str
+  epgs:
+    description:
+    - The name of the EPGs to migrate.
+    type: list
+    elements: dict
```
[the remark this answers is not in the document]
anvitha-jain: Done

plugins/modules/mso_schema_template_migrate.py
lhercot: Not sure they need a query on this module. They can use the query on EPG, BD or event template/schemas.

lhercot (changes requested):

plugins/modules/mso_schema_template_migrate.py
```diff
@@ -0,0 +1,249 @@
...
+        target_schema=dict(type='str', required=True),
+        target_template=dict(type='str', required=True),
+        state=dict(type='str', default='present'),
+    )
+
+    module = AnsibleModule(
+        argument_spec=argument_spec,
+        supports_check_mode=True,
+        required_if=[
+            ['state', 'present', ['target_schema', 'target_template']],
```
lhercot, line 195: I don't think this is required as it is already required=True.

tests/integration/targets/mso_schema_template_migrate/tasks/main.yml
```diff
@@ -0,0 +1,288 @@
...
+    output_level: '{{ mso_output_level | default("info") }}'
+    site: '{{ mso_site | default("ansible_test") }}'
+    apic_username: '{{ apic_username }}'
+    apic_password: '{{ apic_password }}'
+    apic_site_id: '{{ apic_site_id | default(102) }}'
+    urls:
+    - https://{{ apic_hostname }}
+    state: present
+
+#  - name: Remove EPGs in schema 1
```
lhercot, line 52: If we don't use it, remove the commented portions.

lhercot (changes requested):

tests/integration/targets/mso_schema_template_migrate/tasks/main.yml
```diff
@@ -0,0 +1,288 @@
...
+
+- name: Query MSO version
+  mso_version:
+    <<: *mso_info
+    state: query
+  register: version
+
+- name: Set version vars
+  set_fact:
+    mso_l3mcast: false
```
lhercot, line 31: I don't think you use this variable so we can remove this block.

tests/integration/targets/mso_schema_template_migrate/tasks/main.yml
```diff
@@ -0,0 +1,288 @@
...
+  mso_schema_template_bd:
+    <<: *vrf_present
+    schema: '{{ mso_schema | default("ansible_test") }}_2'
+    template: Template 1
+    bd: '{{ item }}'
+    vrf:
+      name: VRF
+      schema: '{{ mso_schema | default("ansible_test") }}'
+      template: Template 1
+    #  layer3_multicast: true
```
lhercot, line 192: remove the commented line

lhercot:
I think you have to skip the migrate test between schemas on the 2.2(4) version because of the L3MCast bug.

lhercot (approved):
LGTM

PR #129, opened 2020-11-13
remove 'present' from mso_schema

### LINKED ISSUE #123: mso_schema module generates an invalid POST request.
When the first example on [this Doc Page](https://docs.ansible.com/ansible/latest/collections/cisco/mso/mso_schema_module.html#parameter-templates) is used. The MSO will return the following error:

```
"msg":"MSO Error 125: Malformed body",
"payload":{
   "code":125,
   "info":{
      "obj[0]":[
         "\"{'displayName': 'template1', 'name': 'template1', 'tenantId': 'north_europe'}\" is not an object"
      ]
   },
   "message":"Malformed body"
}
```

I discussed this problem in [this post](https://groups.google.com/g/ansible-project/c/on64onW7MSU) in the Ansible mailing list, where someone noticed that this bug was introduced in commit [98f010a6889c45536b925a0b77a99e816922929e](https://github.com/CiscoDevNet/ansible-mso/commit/98f010a6889c45536b925a0b77a99e816922929e), by setting the element type of the template list to string instead of dict.

xinyuezhao:
resolve #123 

- Remove 'present' state from mso_schema
- Fix documentation for mso_schema
- Add test file for mso_schema

### DIFF
--- a/plugins/modules/mso_schema.py
+++ b/plugins/modules/mso_schema.py
@@ -31,7 +31,7 @@ options:
     - Use C(query) for listing an object or multiple objects.
     type: str
     choices: [ absent, query ]
-    default: present
+    default: query
 notes:
 - Due to restrictions of the MSO REST API this module cannot create empty schemas (i.e. schemas without templates).
   Use the M(cisco.mso.mso_schema_template) to automatically create schemas with templates.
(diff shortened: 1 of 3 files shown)


lhercot (approved):
LGTM

PR #133, opened 2020-11-20
[mso_schema_template_external_epg] Addition of cloud setting for ext epg fixes #128

### LINKED ISSUE #128: external EPG for public cloud
when creating an external EPG for a public cloud site, the L3out object is not required.
cisco.mso.mso_schema_template_external_epg assumes an L3out exists, but as per the above statement, this is not a requirement for public cloud.
cisco.mso.mso_schema_template_external_epg is missing the following:
- site type, which by default is on-prem. there should be an option to select cloud.
- if the site type is cloud, no L3out should be requested, but an Application Profile.

shrsr:


plugins/modules/mso_schema_template_external_epg.py
```diff
@@ -35,6 +35,13 @@
     - The name of the external EPG to manage.
     type: str
     aliases: [ name, externalepg ]
+  external_epg_type:
```
lhercot, line 38: let's call it "type"

plugins/modules/mso_schema_template_external_epg.py
```diff
@@ -35,6 +35,13 @@
...
+    description:
+    - The type of external epg.
+    - anp needs to be associated with external epg when the type is cloud.
```
lhercot, line 41: add that l3out can be associated with external epg when the type is on-premises.

plugins/modules/mso_schema_template_external_epg.py
```diff
@@ -260,11 +294,13 @@ def main():
             name=external_epg,
             displayName=display_name,
             vrfRef=vrf_ref,
+            extEpgType=external_epg_type,
```
lhercot, line 297: only add to payload if cloud is used. This will prevent breakage in previous MSO versions that don't support it.

plugins/modules/mso_schema_template_external_epg.py
```diff
@@ -260,11 +294,13 @@ def main():
...
             l3outRef=l3out_ref,
```
lhercot, line 298: remove it from here and do a different condition below.

plugins/modules/mso_schema_template_external_epg.py
```diff
@@ -260,11 +294,13 @@ def main():
...
             preferredGroup=preferred_group,
         )
 
-        if anp is not None:
+        if external_epg_type == 'cloud':
+            del payload['l3outRef']
             payload['anpRef'] = anp_ref
```
lhercot, line 304: add else for the on-premises case to avoid the deletion of the key.

tests/integration/targets/mso_schema_template_external_epg/tasks/main.yml
```diff
@@ -428,6 +454,152 @@
...
+    <<: *mso_info
+    schema: '{{ mso_schema | default("ansible_test") }}'
+    template: Template 2
+    external_epg: ansible_test_6
+    external_epg_type: cloud
+    vrf:
+      name: VRF
+      schema: '{{ mso_schema | default("ansible_test") }}'
+      template: Template 1
+    l3out:
```
lhercot, line 468: l3out does not make sense when configuring type = cloud.

tests/integration/targets/mso_schema_template_external_epg/tasks/main.yml
```diff
@@ -428,6 +454,152 @@
...
```
lhercot, line 499: same

tests/integration/targets/mso_schema_template_external_epg/tasks/main.yml
```diff
@@ -428,6 +454,152 @@
...
+    <<: *mso_info
+    schema: '{{ mso_schema | default("ansible_test") }}'
+    template: Template 2
+    external_epg: ansible_test_6
+    external_epg_type: cloud
+    vrf:
+      name: VRF2
+      schema: '{{ mso_schema | default("ansible_test") }}'
+      template: Template 1
+    l3out:
```
lhercot, line 527: same

tests/integration/targets/mso_schema_template_external_epg/tasks/main.yml
```diff
@@ -428,6 +454,152 @@
...
+    external_epg_type: on-premise
+    vrf:
+      name: VRF
+      schema: '{{ mso_schema | default("ansible_test") }}'
+      template: Template 1
+    l3out:
+      name: L3out
+      schema: '{{ mso_schema | default("ansible_test") }}'
+      template: Template 1
+    anp:
```
lhercot, line 561: no ANP here.

tests/integration/targets/mso_schema_template_external_epg/tasks/main.yml
```diff
@@ -428,6 +454,152 @@
...
+    external_epg: ansible_test_7
+    vrf:
+      name: VRF
+      schema: '{{ mso_schema | default("ansible_test") }}'
+      template: Template 1
+    l3out:
+      name: L3out
+      schema: '{{ mso_schema | default("ansible_test") }}'
+      template: Template 1
+    anp:
```
lhercot, line 590: no anp here.

lhercot (changes requested):

plugins/modules/mso_schema_template_external_epg.py
shrsr: if external_epg_type == 'cloud':
            payload['extEpgType'] = type
            payload['anpRef'] = anp_ref
else:

lhercot (approved):

PR #157, opened 2021-01-28
[mso_schema_site_anp_epg_staticport] Made changes to reduce Schema API request size by using list-identity

anvitha-jain:
… respect to schema usage

### DIFF
--- a/plugins/module_utils/mso.py
+++ b/plugins/module_utils/mso.py
@@ -561,12 +561,15 @@ class MSOModule(object):
         if schema is None:
             return schema
 
-        s = self.get_obj('schemas', displayName=schema)
-        if not s:
+        schema_summary = self.query_objs('schemas/list-identity', key='schemas', displayName=schema)
+        if not schema_summary:
+            self.fail_json(msg="Provided schema '{0}' does not exist".format(schema))
+        schema_id = schema_summary[0].get('id')
+        if not schema_id:
             self.module.fail_json(msg="Schema '%s' is not a valid schema name." % schema)
-        if 'id' not in s:
-            self.module.fail_json(msg="Schema lookup failed for schema '%s': %s" % (schema, s))
-        return s.get('id')
+        # if 'id' is not schema_id:
+        #     self.module.fail_json(msg="Schema lookup failed for schema '%s': %s" % (schema, schema_id))
+        return schema_id
 
     def lookup_domain(self, domain):
         ''' Look up a domain and return its id '''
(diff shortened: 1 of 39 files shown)


plugins/module_utils/mso.py
```diff
@@ -947,3 +947,10 @@ def update_filter_obj(self, contract_obj, filter_obj, filter_type, contract_disp
         filter_obj['contractScope'] = contract_obj.get('scope')
         filter_obj['contractFilterType'] = contract_obj.get('filterType')
         return filter_obj
+
+    def get_schema_details(self, schema):
+        schema_summary = self.query_objs('schemas/list-identity', key='schemas', displayName=schema)
+        if not schema_summary:
+            self.fail_json(msg="Provided schema '{0}' does not exist".format(schema))
+        schema_id = schema_summary[0].get('id')
```
lhercot, line 955: What if query_objs return more than one result? Is that possible?

plugins/module_utils/mso.py
```diff
@@ -947,3 +947,10 @@ def update_filter_obj(self, contract_obj, filter_obj, filter_type, contract_disp
         filter_obj['contractScope'] = contract_obj.get('scope')
         filter_obj['contractFilterType'] = contract_obj.get('filterType')
         return filter_obj
+
+    def get_schema_details(self, schema):
```
lhercot, line 951: Is it the details or just the id? Name is probably not correct

plugins/modules/mso_schema_site_anp_epg_staticport.py
```diff
@@ -284,11 +284,9 @@ def main():
     mso = MSOModule(module)
 
     # Get schema_id
-    schema_obj = mso.get_obj('schemas', displayName=schema)
-    if not schema_obj:
-        mso.fail_json(msg="Provided schema '{0}' does not exist".format(schema))
-    schema_path = 'schemas/{id}'.format(**schema_obj)
-    schema_id = schema_obj.get('id')
+    schema_id = mso.get_schema_details(schema)
+    schema_path = "schemas/{0}".format(schema_id)
```
lhercot, line 288: should we just return this as part of the previous method as that path should always be the same.

lhercot (changes requested):

yannickcastano:
Hi team,
Other modules are also using this #Get schema_id portion of code (probably all the mso_schema_*.py). Can you find all the module using it in order to apply the optimisation ?
    # Get schema_id
    schema_obj = mso.get_obj('schemas', displayName=schema)
    if not schema_obj:
        mso.fail_json(msg="Provided schema '{0}' does not exist".format(schema))

    schema_path = 'schemas/{id}'.format(**schema_obj)
    schema_id = schema_obj.get('id')
Thanks,
/Yannick

lhercot:
@yannickcastano Yes, this is just the first review. Wanted to check it before we apply it across the board.

Issue #156, opened 2021-01-24
Reduce Schema API request size by using "list-identity"


lhercot:
Every schema module query all the schemas with all their configurations which can create issue to users with large datasets.

MSO API provides the /api/v1/schemas/list-identity endpoint to only list name, IDs and templates of schemas in the list reducing the size of the response.

All MSO modules currently querying /schemas should adopt this method (except maybe the mso_schema module when state query).

PR #158, opened 2021-02-02
[mso] Made changes to avoid recurrent use of schema objects in modules. issue #156

anvitha-jain:
This solves issue #156 with the use of "list-identity" to reduce the size by using Schema API request.

### DIFF
--- a/plugins/modules/mso_schema_site_anp.py
+++ b/plugins/modules/mso_schema_site_anp.py
@@ -139,13 +139,9 @@ def main():
 
     mso = MSOModule(module)
 
-    # Get schema_id
-    schema_obj = mso.get_obj('schemas', displayName=schema)
-    if not schema_obj:
-        mso.fail_json(msg="Provided schema '{0}' does not exist".format(schema))
-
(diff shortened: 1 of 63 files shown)


plugins/modules/mso_schema_site_anp.py
```diff
@@ -139,13 +139,9 @@ def main():
...
-    # Get schema_id
-    schema_obj = mso.get_obj('schemas', displayName=schema)
-    if not schema_obj:
-        mso.fail_json(msg="Provided schema '{0}' does not exist".format(schema))
-
-    schema_path = 'schemas/{id}'.format(**schema_obj)
-    schema_id = schema_obj.get('id')
+    # Get schema objects
+    schema_id = mso.lookup_schema(schema)
+    schema_path, schema_obj = mso.query_schema(schema)
```
lhercot, line 144: Any reason not to return schema_id as part of query_schema?
This way you are making 2 times a query to the list of schemas where you could make only one.

lhercot (changes requested):

lhercot (approved):
LGTM

PR #164, opened 2021-03-10
Fix Ansible auth issue with ND/MSO 3.2

mminnen:
Cisco MSO 3.2 in combination with Nexus Dashboard responds with a 200 OK instead of 201 OK. This fixes the authentication failed error from Ansible with MSO 3.2.


lhercot:
I suppose you also changed the login URL to make it work with MSO 3.2 on ND, am I correct?

lhercot (approved):

lhercot:
I don't think this fix will hold as the login URL is being deprecated. Proper fix is coming soon but this does not hurt so merging.

PR #188, opened 2021-05-21
Test for same template names in different schemas and tenants

shrsr:
Addition of test cases

### DIFF
--- a/tests/integration/targets/mso_schema_template/tasks/main.yml
+++ b/tests/integration/targets/mso_schema_template/tasks/main.yml
@@ -38,18 +38,22 @@
     schema: '{{ item }}'
     state: absent
   loop:
+  - '{{ mso_schema | default("ansible_test") }}_3'
   - '{{ mso_schema | default("ansible_test") }}_2'
   - '{{ mso_schema | default("ansible_test") }}'
 
 - name: Ensure tenant ansible_test exists
   mso_tenant: 
     <<: *mso_info
-    tenant: ansible_test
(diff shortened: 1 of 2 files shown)


tests/integration/targets/mso_schema_template/tasks/main.yml
```diff
@@ -38,18 +38,22 @@
...
-    tenant: ansible_test
+    tenant: '{{ item }}'
     users:
       - '{{ mso_username }}'
     sites:
     - '{{ mso_site | default("ansible_test") }}'
     state: present
+  loop:
+  - 'ansible_test'
+  - 'ansible_test_2'
```
lhercot, line 56: You add a tenant here and a schema after but does not change the cleanup section of the next module so I am afraid one module will fail at the next execution of this one.

lhercot (changes requested):

tests/integration/targets/mso_schema_template/tasks/main.yml
shrsr: I added a cleanup section for the extra tenant and schema I included here. That should take care of this right?

lhercot: You might need to add those at the top of the next module also in case this module fails so the future modules don't have any issues.

shrsr: okay

lhercot (approved):
LGTM

Issue #190, opened 2021-05-24
Add support for Schema and Template Cloning (MSO 3.3)


lhercot:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Create new modules to clone Schema's or Template's based on new feature of MSO 3.3

### New or Affected Module(s):

* mso_schema_clone
* mso_schema_template_clone

### MSO version and MSO Platform

* V3.3.x ND-based.

### APIC version and APIC Platform for Site Level Resources

* N/A

PR #196, opened 2021-05-25
[mso_schema_clone] Supports cloning the schema issue #190

anvitha-jain:
solves #190

### DIFF
--- a/plugins/modules/mso_schema_clone.py
+++ b/plugins/modules/mso_schema_clone.py
@@ -1,0 +1,118 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# Copyright: (c) 2021, Anvitha Jain (@anvitha-jain) <anvjain@cisco.com>
+# GNU General Public License v3.0+ (see LICENSE or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+ANSIBLE_METADATA = {'metadata_version': '1.1',
+                    'status': ['preview'],
+                    'supported_by': 'community'}
+
+DOCUMENTATION = r'''
+---
+module: mso_schema_clone
+short_description: Clone schemas
+description:
+- Clone schemas on Cisco ACI Multi-Site.
+- Clones only template objects and not site objects.
+- This module can only be used on versions of MSO that are 3.3 or greater.
+author:
+- Anvitha Jain (@anvitha-jain)
+options:
+  source_schema:
+    description:
+    - The name of the source_schema.
+    type: str
+  destination_schema:
+    description:
+    - The name of the destination_schema.
+    type: str
+    aliases: [ name, displayName ]
+  state:
+    description:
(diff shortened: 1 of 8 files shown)


plugins/modules/mso_schema_clone.py
```diff
@@ -0,0 +1,118 @@
...
+    type: str
+    aliases: [ name, displayName ]
+  state:
+    description:
+    - Use C(clone) for adding.
+    type: str
+    choices: [ clone ]
+    default: clone
+notes:
+- Due to restrictions of the MSO REST API this module cannot create empty schemas (i.e. schemas without templates).
```
lhercot, line 41: I don't think this note is needed here.

plugins/modules/mso_schema_clone.py
```diff
@@ -0,0 +1,118 @@
...
+  Use the M(cisco.mso.mso_schema_template) to automatically create schemas with templates.
```
lhercot, line 42: Same here

plugins/modules/mso_schema_clone.py
```diff
@@ -0,0 +1,118 @@
...
+options:
+  source_schema:
+    description:
+    - The name of the source_schema.
+    type: str
+  destination_schema:
+    description:
+    - The name of the destination_schema.
+    type: str
+    aliases: [ name, displayName ]
```
lhercot, line 33: remove aliases. I don't think they are clear enough.

plugins/modules/mso_schema_clone.py
```diff
@@ -0,0 +1,118 @@
...
+    destination_schema_id = None
+    path = 'schemas'
+
+    # Check if source and destination schema are named differently
+    if source_schema == destination_schema:
+        mso.fail_json(msg="Schema with the name '{0}' already exists. Please use another name.".format(destination_schema))
+    # Query for existing object(s)
+    if destination_schema:
+        mso.existing = mso.get_obj(path, displayName=destination_schema)
+        if mso.existing:
```
lhercot, line 96: Does the UI allow you to clone to a destination schema that already exists?

plugins/modules/mso_schema_clone.py
```diff
@@ -0,0 +1,118 @@
...
+    destination_schema = module.params.get('destination_schema')
+    state = module.params.get('state')
+
+    mso = MSOModule(module)
+
+    # Get source schema details
+    source_schema_path = "schemas/{0}".format(mso.lookup_schema(source_schema))
+    source_schema_obj = mso.query_obj(source_schema_path, displayName=source_schema)
+
+    source_data = source_schema_obj.get('templates')
```
lhercot, line 85: You don't seems to fail if source_schema does not exist. Is that normal?

lhercot (changes requested):

plugins/modules/mso_schema_clone.py
anvitha-jain: I check this condition in the line above.

lhercot (approved):

PR #222, opened 2021-12-11
Adding support to the DCNM

cassiolange:
adding two new modules, updating the mso_schema_template to support Network templates (DCNM) and  fixing the user and site lookup.


anvitha-jain (changes requested):
README.md file changes are not necessary. There is no test case file to test the modules created.

PR #268, opened 2022-05-26
Added additional validations and implemented logic to delete the tenant from MSO and Sites

sajagana:
Note:

Added additional validations to the mso_tenant module and implemented logic to delete the tenant from MSO and Sites.

### DIFF
--- a/plugins/modules/mso_tenant.py
+++ b/plugins/modules/mso_tenant.py
@@ -120,6 +120,7 @@ RETURN = r"""
 
 from ansible.module_utils.basic import AnsibleModule
 from ansible_collections.cisco.mso.plugins.module_utils.mso import MSOModule, mso_argument_spec
+from ansible_collections.cisco.mso.plugins.module_utils.constants import YES_OR_NO_TO_BOOL_STRING_MAP
 
 
 def main():
@@ -177,8 +178,7 @@ def main():
             if module.check_mode:
                 mso.existing = {}
             else:
-                orchestrator_only_map = {"yes": "true", "no": "false"}
-                path = "{0}?msc-only={1}".format(path, orchestrator_only_map.get(orchestrator_only))
+                path = "{0}?msc-only={1}".format(path, YES_OR_NO_TO_BOOL_STRING_MAP.get(orchestrator_only))
                 mso.existing = mso.request(path, method="DELETE")
 
     elif state == "present":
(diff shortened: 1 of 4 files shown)


plugins/modules/mso_tenant.py
```diff
@@ -108,59 +125,63 @@
...
 
-    elif state == 'absent':
+    elif state == "absent":
         mso.previous = mso.existing
         if mso.existing:
             if module.check_mode:
                 mso.existing = {}
             else:
-                mso.existing = mso.request(path, method='DELETE')
+                orchestrator_only_map = {"yes": "true", "no": "false"}
```
akinross, line 180: Should we define this as a generic constant for toggling true/false to yes/no and other way around? I did create a constants file where we could place these. 

https://github.com/CiscoDevNet/ansible-mso/blob/master/plugins/module_utils/constants.py

plugins/modules/mso_tenant.py
```diff
@@ -6,13 +6,12 @@
...
+
 __metaclass__ = type
 
-ANSIBLE_METADATA = {'metadata_version': '1.1',
-                    'status': ['preview'],
-                    'supported_by': 'community'}
+ANSIBLE_METADATA = {"metadata_version": "1.1", "status": ["preview"], "supported_by": "community"}
 
-DOCUMENTATION = r'''
+DOCUMENTATION = r"""
```
akinross, line 14: I see a lot of difference usage of double vs single quotes throughout the other modules. What is our defined standard for this?

sajagana: I am using **black** Python code formatter to align the code. It would be better if we use **black** code formatter.

FYI: https://pypi.org/project/black/

akinross: Ok will we use this then on the entire project directory? 
Or just when we make changes to a module and or piece of code?

sajagana: We can use it for the entire project directory. So that we can avoid formatting errors.

lhercot:
For some reason, tests don't pass on lh-dmz1-pod1-mso-v224 setup. Can you check why?

lhercot (commented):
For some reason, tests don't pass on lh-dmz1-pod1-mso-v224 setup. Can you check why?

sajagana:
173.36.219.25 APIC was not up and couldn’t ping from the local system, which may cause the issue.

lhercot (approved):
LGTM

PR #272, opened 2022-06-02
Added patch_operation attribute to the mso module utility 

sajagana:
Note:

Added patch_operation attribute to the mso module to print the full PATCH operations when output_level is debug and the invocation method is PATCH.


lhercot (approved):
LGTM

PR #285, opened 2022-09-15
Added support for the NDO backup feature

sajagana:
Note:

Added support for the NDO backup feature and validation tasks


anvitha-jain (commented):
Rebase on the latest master.

anvitha-jain (approved):
LGTM

akinross (dismissed):

plugins/module_utils/mso.py
```diff
@@ -294,96 +327,119 @@ def __init__(self, module):
...
-                self.params['use_proxy'] = True
-            if self.params.get('validate_certs') is None:
-                self.params['validate_certs'] = True
+        # Ensure protocol is set
+        self.params["protocol"] = "https" if self.params.get("use_ssl", True) else "http"
 
-            # Ensure protocol is set
-            self.params['protocol'] = 'https' if self.params.get('use_ssl', True) else 'http'
+        if self.params.get("host") is None:
+            self.fail_json(msg="Parameter 'host' is required when not using the HTTP API connection plugin")
```
lhercot, line 345: You can't move this outside of "if self.module._socket_path is None:"  as this only applies to non-HTTPAPI plugin cases.

plugins/module_utils/mso.py
```diff
@@ -294,96 +327,119 @@ def __init__(self, module):
...
                 self.platform = "nd"
 
+            # Set base_uri
+            self.baseuri = "{0}mso/api/v1/".format(self.base_only_uri)
+
+        if self.params.get("password"):
+            # Perform password-based authentication, log on using password
+            self.login()
+        else:
+            self.fail_json(msg="Parameter 'password' is required for authentication")
```
lhercot, line 377: You can't move this outside of "if self.module._socket_path is None:" as this only applies to non-HTTPAPI plugin cases.

plugins/module_utils/mso.py
```diff
@@ -294,96 +327,119 @@ def __init__(self, module):
...
-                               use_proxy=self.params.get('use_proxy'))
+            payload = {
+                "username": self.params.get("username", "admin"),
+                "password": self.params.get("password"),
+            }
+
+        self.url = urljoin(self.baseuri, "auth/login")
+
+        if self.platform == "nd":
+            self.url = "{0}login".format(self.base_only_uri)
```
lhercot, line 410: This login function shouldn't be used for ND so not sure why we are adding this here.

plugins/module_utils/mso.py
```diff
@@ -453,35 +521,37 @@ def request_download(self, path, destination=None):
...
     def request_upload(self, path, fields=None):
-        ''' Generic HTTP MultiPart POST method for MSO uploads. '''
+        """Generic HTTP MultiPart POST method for MSO uploads."""
         self.path = path
         self.url = urljoin(self.baseuri, path)
 
-        if not HAS_MULTIPART_ENCODER:
-            self.fail_json(msg='requests-toolbelt is required for the upload state of this module')
+        # if not HAS_MULTIPART_ENCODER:
+        #     self.fail_json(msg='requests-toolbelt is required for the upload state of this module HAS_MULTIPART_ENCODER is {}'.format(HAS_MULTIPART_ENCODER))
```
lhercot, line 529: Why commenting this?

tests/integration/targets/mso_backup/tasks/main.yml
```diff
@@ -49,464 +64,814 @@
...
+
+    - name: Verify query_non_existent_backup
+      assert:
+        that:
+          - query_non_existent_backup is not changed
+
+# Backup validation tasks for MSO/NDO version > 3.2
+- name: Execute tasks only for MSO version > 3.2
+  when: version.current.version is version('3.2', '>')
+  block:
```
lhercot, line 509: Why do we need a complete separate one for v3.2+ most of the existing task above could apply for both right? (if you improved them we could run your new tasks for both too)

lhercot (changes requested):

Issue #299, opened 2022-11-14
mso_schema_template_deploy doesn't work


angelogemma:
Hi Team,
I have some issue with the module used to deploy a template on MSO Servicie of Nexus Dashboard Orchestrator Version 2.2(2d).

This is the excerpt of the ansible playbook:
    - name: DEPLOY DEL TEMPLATE
      cisco.mso.mso_schema_template_deploy:
        host: "{{ inventory_hostname }}"
        username: "{{ username }}"
        password: "{{ password }}"
        validate_certs: False
        schema: "SCHEMA1"
        template: "TEMP1"
        state: deploy

This is the returned error from ansible:
`TASK [DEPLOY DEL TEMPLATE] *******************************************************************************************************************************************************************
fatal: [10.10.20.60]: FAILED! => {"changed": false, "current": {}, "msg": "Connection failed for https://10.10.20.60:443/mso/api/v1/execute/schema/637270235aa848e4be8e6901/template/TEMP1. OK (19 bytes)", "status": 404}`

When I deploy a template from a python script I use the following and it works.

POST : https://10.10.20.60:443/mso/api/v1/task
Body: {"schemaId":"6372296221ecac1dc5370ce1","templateName":"TEMP4","isRedeploy":true,"_updateVersion":0}

Please can you check?
Thank you in advance.

akinross:
Hi @angelogemma, I think you provided the Nexus Dashboard version, what version of NDO are you currently running?

angelogemma:
Yes @akinross , you are right. Sorry. The MSO version is 4.0.2i.
I was using the sandbox provided by Cisco DevNet.
Thank you very much!
Angelo

akinross:
Hi @angelogemma, the api endpoint for deploy has changed in NDO4.0 so this module will only provide the functionality for NDO versions before 4.0. We are working on implementing a new module for this new api endpoint.

angelogemma:
Thank you @akinross .
Have a good day.

angelogemma:
In the meantime, if someone needs a workaround, in this way it works:
    - name: FIND SCHEMA ID
      cisco.mso.mso_schema:
        host: "{{ inventory_hostname }}"
        username: "{{ username }}"
        password: "{{ password }}"
        validate_certs: False
        schema: "{{ schema }}"
        state: "query"
      register: schemaID

    - name: DEPLOY TEMPLATE
      cisco.mso.mso_rest:
[...]

PR #305, opened 2022-11-29
[minor_change] ndo 4+ deploy functionality

akinross:
fix for #299


plugins/modules/ndo_schema_template_deploy.py
```diff
@@ -0,0 +1,139 @@
...
+  sites:
+    description:
+    - The name of the site(s).
+    type: list
+    elements: str
+  state:
+    description:
+    - Use C(deploy) to deploy schema template.
+    - Use C(redeploy) to redeploy schema template.
+    - Use C(undeploy) to deploy schema template from a site.
```
sajagana, line 42: undeploy is used to delete the object or for any other purpose?

plugins/modules/ndo_schema_template_deploy.py
```diff
@@ -0,0 +1,139 @@
...
+    - Use C(query) to get deployment status.
+    type: str
+    choices: [ deploy, redeploy, undeploy, query ]
```
sajagana, line 45: Which option is used to delete the schema template?

plugins/modules/ndo_schema_template_deploy.py
```diff
@@ -0,0 +1,139 @@
...
+  cisco.mso.ndo_schema_template_deploy:
+    host: mso_host
+    username: admin
+    password: SomeSecretPassword
+    schema: Schema 1
+    template: Template 1
+    sites: [ Site1, Site2 ]
+    state: undeploy
+  delegate_to: localhost
+'''
```
sajagana, line 84: You can add examples to query single and multiple objects.

plugins/modules/ndo_schema_template_deploy.py
```diff
@@ -0,0 +1,139 @@
...
+
+from ansible.module_utils.basic import AnsibleModule
+from ansible_collections.cisco.mso.plugins.module_utils.mso import MSOModule, mso_argument_spec
+
+
+def main():
+    argument_spec = mso_argument_spec()
+    argument_spec.update(
+        schema=dict(type="str", required=True),
+        template=dict(type="str", required=True),
```
sajagana, line 97: You can include schema and template under the required_if.
It may help to query all template objects without specifying the schema and template name.

sajagana (commented):
You can try adding Idempotency tasks.

plugins/modules/ndo_schema_template_deploy.py
akinross: to undeploy a template, see there is a typo will fix

plugins/modules/ndo_schema_template_deploy.py
akinross: you don't delete a template but undeploy it from a site thus the state differ from other modules

plugins/modules/ndo_schema_template_deploy.py
akinross: not sure what you mean, but assume you want to see query result of all deploy status? can check it the api allows, else schema and template are required to retrieve the status

plugins/modules/ndo_schema_template_deploy.py
sajagana: Thank you Akini!

plugins/modules/ndo_schema_template_deploy.py
akinross: could not query status generic without schema/template so these are required.

tests/integration/inventory.networking
```diff
@@ -3,9 +3,14 @@ lh-dmz1-pod1-mso-v224 ansible_host=173.36.219.29 ansible_connection=local mso_ho
 lh-dmz1-pod1-mso-v311 ansible_host=173.36.219.11 ansible_connection=local mso_hostname=173.36.219.11
 lh-dmz1-pod1-ndo-v371 ansible_host=173.36.219.30 ansible_connection=ansible.netcommon.httpapi ansible_network_os=cisco.nd.nd mso_hostname=173.36.219.30
 
-[mso:vars]
+[ndo]
+lh-dmz1-pod1-ndo-v400 ansible_host=173.36.219.32 ansible_connection=ansible.netcommon.httpapi ansible_network_os=cisco.nd.nd ndo_hostname=173.36.219.32
+
+[all:vars]
```
anvitha-jain, line 9: doesn't this affect the usage of other modules?

anvitha-jain (changes requested):

tests/integration/inventory.networking
akinross: I have tested on other module for it to still work, this is also why I renamed mso:vars to all:vars. Due to the prepend change of the new module I needed to make a change to this to get it working. We can discuss alternative methods if you have a better approach.

anvitha-jain (approved):
LGTM

lhercot (approved):
LGTM

PR #316, opened 2023-01-23
[ignore] Update ansible and action versions used in tests

lhercot:


.github/workflows/ansible-test.yml
```diff
@@ -11,19 +11,20 @@ jobs:
...
     steps:
     - name: Check out code
       uses: actions/checkout@v3
     
-    - name: Set up Python 3.8
+    - name: Set up Python 3.9
       uses: actions/setup-python@v4
       with:
-        python-version: 3.8
+        python-version: 3.9
```
akinross, line 24: should/can we make this a variable so we can do easier change of python version next time we bump to 3.10?

.github/workflows/ansible-test.yml
```diff
@@ -149,42 +173,41 @@ jobs:
...
     steps:
-    - name: Set up Python 3.8
+    - name: Set up Python 3.9
       uses: actions/setup-python@v4
       with:
-        python-version: 3.8
-  
+        python-version: 3.9
+
     - name: Install ansible-base (v2.11.12)
```
akinross, line 199: is this correct version or should we do higher version for integration testing?

akinross: also because in line 66 we install a higher version.

akinross (approved):
LGTM

PR #317, opened 2023-01-25
[bugfix] Changes made to the mso plugin to query domain ids when the domain name is provided

shrsr:


plugins/httpapi/mso.py
```diff
@@ -236,33 +236,38 @@ def _response_to_json(self, response_data):
...
+                    if domain.get("name") == domain_name:
                         if "id" in domain:
                             return domain.get("id")
                         else:
-                            self.error = dict(code=-1, message="Login domain '{0}' is not a valid domain name.".format(domain))
-                            raise ConnectionError(self._verify_response(None, None, None, None))
-                self.error = dict(code=-1, message="Login domain lookup failed for domain '{0}': {1}".format(domain_name, domain))
-                raise ConnectionError(self._verify_response(None, None, None, None))
+                            self.error = dict(code=self.status, message="Login domain lookup failed for domain '{0}': {1}".format(domain_name, domain))
+                            raise ConnectionError(json.dumps(self._verify_response(response, method, full_path, data)))
```
akinross, line 265: https://github.com/CiscoDevNet/ansible-mso/pull/317/files#diff-4a71c6ecf7c1f1180c64c9c2157b5c305a3c2b9c3d78a39ce1c5a96f349ab44cL206

what is the response code in this method "_verify_response" doing and why is this not set to self.status? also what isn the purpose of this code setting in this function?

plugins/httpapi/mso.py
```diff
@@ -236,33 +236,38 @@ def _response_to_json(self, response_data):
...
+                self.error = dict(code=self.status, message="Login domain '{0}' is not a valid domain name.".format(domain_name))
+                raise ConnectionError(json.dumps(self._verify_response(response, method, full_path, data)))
             else:
-                self.error = dict(code=-1, message="Key 'domains' missing from data")
-                raise ConnectionError(self._verify_response(None, None, None, None))
+                self.error = dict(code=self.status, message="Key 'domains' missing from data")
+                raise ConnectionError(json.dumps(self._verify_response(response, method, full_path, data)))
```
akinross, line 270: would it be an option to have less copy/paste code? for example by setting error_message to the string and moving the error and raise to end of the if? something like example below (have not tested it)

```
error_message = ""
if data is not None:
    response_data = self._response_to_json(data)
    domains = response_data.get("domains")
    if domains is not None:
        for domain in domains:
            if domain.get("name") == domain_name:
                if "id" in domain:
                    return domain.get("id")
                else:
                    error_message = "Login domain lookup failed for domain '{0}': {1}".format(domain_name, domain)
        error_message = "Login domain '{0}' is not a valid domain name.".format(domain_name)
    else:
        error_message = "Key 'domains' missing from data"

if error_message:
    self.error = dict(code=self.status, message=error_message)
    raise ConnectionError(json.dumps(self._verify_response(response, method, full_path, data)))
```

plugins/httpapi/mso.py
```diff
@@ -236,33 +236,38 @@ def _response_to_json(self, response_data):
...
+
+        method = "GET"
+        path = "/mso/api/v1/auth/login-domains"
+        full_path = self.connection.get_option("host") + path
+
         # TODO: Replace response by -
-        response, data = self.send_request("GET", "auth/login-domains")
+        response, data = self.connection.send(path, None, method=method, headers=self.headers)
 
         if data is not None:
```
akinross, line 255: only during not None? is there option of [] return? what would then be behaviour? error?

plugins/httpapi/mso.py
lhercot: There is a difference between status and error code. I don't think we should change all those error codes to self.status. We had used -1 as error code so we know it is not related to a code from MSO nor an HTTP error status as those error are either decoding, missing data or other.

lhercot (changes requested):

plugins/httpapi/mso.py
```diff
@@ -236,33 +236,38 @@ def _response_to_json(self, response_data):
         # JSONDecodeError only available on Python 3.5+
         except Exception as e:
             # Expose RAW output for troubleshooting
-            self.error = dict(code=-1, message="Unable to parse output as JSON, see 'raw' output. {0}".format(e))
+            self.error = dict(code=self.status, message="Unable to parse output as JSON, see 'raw' output. {0}".format(e))
```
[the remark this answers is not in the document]
shrsr: I missed reading the snippet self.status = response.getcode(). I'll revert it.

plugins/httpapi/mso.py
shrsr: Irrespective of what is returned, the general idea is make an attempt to parse the data as JSON. The _response_to_json() function takes care of any exceptions that may be encountered.

plugins/httpapi/mso.py
shrsr: It goes both ways. By doing the above we increase the number of lines of code. Let me know if you still want to change it.

plugins/httpapi/mso.py
akinross: ok but if data is [] it is valid data and would do execution of _response_to_json() which is not needed since it is empty list? would if data be sufficient? an error should still be raised in that case I think.

lhercot (approved):
LGTM

PR #318, opened 2023-01-25
[ignore] removing when statements for ndo 4

akinross:


plugins/modules/ndo_schema_template_deploy.py
```diff
@@ -134,8 +134,10 @@ def main():
         method = "POST"
         payload = dict(schemaId=schema_id, templateName=template)
         if state == "deploy":
+            mso.request("schemas/{id}/validate".format(id=schema_id), method="GET")
```
lhercot, line 137: Do you not need to check the return and error out?

plugins/modules/ndo_schema_template_deploy.py
```diff
@@ -134,8 +134,10 @@ def main():
...
             payload.update(isRedeploy=False)
         elif state == "redeploy":
+            mso.request("schemas/{id}/validate".format(id=schema_id), method="GET")
```
lhercot, line 140: Do you not need to check the return and error out?

tests/integration/targets/mso_schema_site_l3out/tasks/main.yml
```diff
@@ -100,14 +101,14 @@
...
+    l3out: '{{ item }}'
     vrf:
         name: VRF1
     state: present
   register: add_l3out
   loop:
-  - { l3out: L3out1}
-  - { l3out: L3out2}
+  - "{{ l3outs[0] }}"
+  - "{{ l3outs[1] }}"
```
lhercot, line 111:
```
loop: {{ l3outs }}
```
Isn't this enough?

tests/integration/targets/mso_schema_template_contract_filter/tasks/main.yml
```diff
@@ -74,62 +74,62 @@
     template: Template 3
     state: present
 
-- name: Ensure Filter 1 exist
+- name: Ensure Filter1 exist
```
lhercot, line 77: Are space not allowed anymore?

lhercot (changes requested):

plugins/modules/ndo_schema_template_deploy.py
akinross: no it will error out in the request module, wrote tests for this also which checks the error

plugins/modules/ndo_schema_template_deploy.py
akinross: no it will error out ( call fail_json() ) in the request module, wrote tests for this also which checks the error

tests/integration/targets/mso_schema_site_l3out/tasks/main.yml
akinross: will give it another try, didn't have success with similar attempt to do this as it was passing in the entire list and putting that as the name of the l3out

tests/integration/targets/mso_schema_template_contract_filter/tasks/main.yml
akinross: correct for filter/entry/contract a validation rule is triggered by NDO

tests/integration/targets/mso_schema_site_l3out/tasks/main.yml
akinross: see I had '' which messed up the loop with just l3outs, will change to loop: {{ l3outs }}

lhercot (approved):
LGTM

PR #335, opened 2023-03-01
Added route_reachability argument to site_external_epg module

anvitha-jain:


plugins/modules/mso_schema_site_external_epg.py
```diff
@@ -122,8 +171,12 @@ def main():
     if not schema_obj.get("sites"):
         mso.fail_json(msg="No site associated with template '{0}'. Associate the site with the template using mso_schema_site.".format(template))
     sites = [(s.get("siteId"), s.get("templateName")) for s in schema_obj.get("sites")]
+    sites_list = [s.get("siteId") + "/" + s.get("templateName") for s in schema_obj.get("sites")]
```
akinross, line 174: do we need sites_list outside of the scope of the if statement below?

plugins/modules/mso_schema_site_external_epg.py
```diff
@@ -160,7 +214,17 @@ def main():
             ops.append(dict(op="remove", path=op_path))
 
     elif state == "present":
-        l3out_dn = "uni/tn-{0}/out-{1}".format(tenant_name, l3out)
+        # Get external EPGs type from template level
+        external_epgs = [e.get("name") for e in schema_obj.get("templates")[template_idx]["externalEpgs"]]
```
akinross, line 218: do we need this list when external_epg is None? should we move logic inside if statement?

plugins/modules/mso_schema_site_external_epg.py
```diff
@@ -160,7 +214,17 @@ def main():
...
+        if external_epg is not None and external_epg in external_epgs:
```
akinross, line 219: how can external_epg be None during present? Think this part of the if statement is not needed because of the required_if block `["state", "present", ["external_epg"]`

plugins/modules/mso_schema_site_external_epg.py
akinross, line 218: would it be option to get the external epg that matches the name instead of returning a list and indexing? This way I think you can simplify the code inside the if statement.

tests/integration/targets/mso_schema_site_external_epg/tasks/main.yml
```diff
@@ -19,7 +19,7 @@
       validate_certs: '{{ mso_validate_certs | default(false) }}'
       use_ssl: '{{ mso_use_ssl | default(true) }}'
       use_proxy: '{{ mso_use_proxy | default(true) }}'
-      output_level: '{{ mso_output_level | default("info") }}'
+      output_level: '{{ mso_output_level | default("debug") }}'
```
akinross, line 22: reset to info?

tests/integration/targets/mso_schema_site_external_epg/tasks/main.yml
```diff
@@ -67,7 +67,7 @@
     sites:
     - '{{ mso_site | default("ansible_test") }}'
     state: present
-  ignore_errors: yes
+  ignore_errors: true
```
akinross, line 70: what is the reason for the ignore_errors: true change?

plugins/modules/mso_schema_site_external_epg.py
```diff
@@ -160,7 +214,17 @@ def main():
...
+            external_epg_idx = external_epgs.index(external_epg)
+            ext_epg_type = schema_obj.get("templates")[template_idx]["externalEpgs"][external_epg_idx].get("extEpgType")
+            if ext_epg_type != "cloud":
+                if l3out is not None:
+                    l3out_dn = "uni/tn-{0}/out-{1}".format(tenant_name, l3out)
+                else:
+                    mso.fail_json(msg="L3Out cannot be empty when template external EPG type is 'on-premise'.")
```
akinross, line 226: Should we use static 'on-premise' or also considered using the ext_epg_type with formatted string?

tests/integration/targets/mso_schema_site_external_epg/tasks/main.yml
```diff
@@ -666,15 +683,15 @@
...
+    ignore_errors: true
     register: nm_non_existing_site_template
 
   - name: Verify cm_non_existing_site_template and nm_non_existing_site_template
     assert:
       that:
       - cm_non_existing_site_template is not changed
       - nm_non_existing_site_template is not changed
-      - cm_non_existing_site_template.msg == nm_non_existing_site_template.msg == "Provided template 'Template3' does not exist. Existing templates{{':'}} Template1, Template2, Template3"
+      - cm_non_existing_site_template.msg == nm_non_existing_site_template.msg == "Provided site/siteId/template 'ansible_test/6398a4b9f1ca793733c7d609/Template3' does not exist. Existing siteIds/templates{{':'}} 6398a4b9f1ca793733c7d609/Template1, 6398a4b9f1ca793733c7d609/Template2"
```
akinross, line 694: Should we make 6398a4b9f1ca793733c7d609 (siteId) a dynamic value? I assume if sites are reset for some reason that these sites might have new id's which would then break CI.

plugins/modules/mso_schema_site_external_epg.py
```diff
@@ -122,8 +171,12 @@ def main():
     if not schema_obj.get("sites"):
         mso.fail_json(msg="No site associated with template '{0}'. Associate the site with the template using mso_schema_site.".format(template))
     sites = [(s.get("siteId"), s.get("templateName")) for s in schema_obj.get("sites")]
+    sites_list = [s.get("siteId") + "/" + s.get("templateName") for s in schema_obj.get("sites")]
     if (site_id, template) not in sites:
-        mso.fail_json(msg="Provided template '{0}' does not exist. Existing templates: {1}".format(template, ", ".join(templates)))
+        mso.fail_json(
+            msg="Provided site/siteId/template '{0}/{1}/{2}' does not exist. "
```
akinross, line 177: Do we need siteId in the error message? I assume ( which might be wrong ) that the Id would will not mean much to the user ( because the user does not provide the Id but is computed/lookedup from name ).

akinross (changes requested):
See comments.

plugins/modules/mso_schema_site_external_epg.py
anvitha-jain: static on-premise.

akinross (approved):
LGTM

akinross (dismissed):
LGTM

plugins/modules/mso_schema_site_external_epg.py
```diff
@@ -57,13 +65,59 @@
...
+    username: admin
+    password: SomeSecretPassword
+    schema: Schema 1
+    template: Template 1
+    external_epg: External EPG 1
+    l3out: L3out1
+    state: query
+  delegate_to: localhost
+
+- name: Query all Site External EPGs
```
shrsr, line 104: Does this display the site external epgs when the l3out is not set?

shrsr (changes requested):

lhercot (approved):
LGTM

plugins/modules/mso_schema_site_external_epg.py
lhercot: Yes

PR #338, opened 2023-03-08
Made static_ports argument reqired for creation in bulk_static_port module

anvitha-jain:


plugins/module_utils/schema.py
```diff
@@ -112,6 +112,21 @@ def set_template_anp_epg(self, epg, fail_module=True):
...
+        :return: Template anp item. -> Item(Int, Dict) | None
+        """
+        self.validate_schema_objects_present(["template"])
+        kv_list = [KVPair("name", external_epg)]
+        match, existing = self.get_object_from_list(self.schema_objects["template"].details.get("externalEpgs"), kv_list)
+        if not match and fail_module:
+            msg = "Provided External EPG '{0}' not matching existing external_epg(s): {1}".format(external_epg, ", ".join(existing))
+            self.mso.fail_json(msg=msg)
+        self.schema_objects["template_external_epg"] = match
+
```
lhercot, line 129: I don't think this is part of this PR

plugins/modules/mso_schema_site_external_epg.py
```diff
@@ -33,6 +33,7 @@
   l3out:
```
lhercot, line 33: I don't think those changes should be in this PR.

tests/integration/targets/mso_schema_site_external_epg/tasks/main.yml
```diff
@@ -47,16 +47,16 @@
     - https://{{ apic_hostname }}
```
lhercot, line 47: I don't think those changes should be in this PR.

lhercot (changes requested):

PR #368, opened 2023-05-08
Fix for the lookup_site function on the utils mso.py

sajagana:
Note:

Added warning message to the lookup_site function on the utils mso.py.

### DIFF
--- a/plugins/module_utils/mso.py
+++ b/plugins/module_utils/mso.py
@@ -833,7 +833,8 @@ class MSOModule(object):
 
         s = self.get_obj("sites", name=site)
         if not s:
-            self.fail_json(msg="Site '%s' is not a valid site name." % site)
+            self.module.warn("Site '{0}' is not a valid site name.".format(site))
+            return None
         if "id" not in s:
             self.fail_json(msg="Site lookup failed for site '%s': %s" % (site, s))
         return s.get("id")
--- a/plugins/modules/mso_tenant_site.py
+++ b/plugins/modules/mso_tenant_site.py
@@ -261,6 +261,8 @@ def main():
     # Get tenant_id and site_id
     tenant_id = mso.lookup_tenant(module.params.get("tenant"))
     site_id = mso.lookup_site(module.params.get("site"))
+    if not site_id:
+        mso.exit_json()
     tenants = [(t.get("id")) for t in mso.query_objs("tenants")]
     tenant_idx = tenants.index((tenant_id))
 
--- a/tests/integration/targets/mso_tenant_site/tasks/main.yml
+++ b/tests/integration/targets/mso_tenant_site/tasks/main.yml
@@ -45,10 +45,21 @@
     site: '{{ item }}'
     state: absent
   loop:
+  - 'invalid_site_{{ mso_site | default("ansible_test") }}'
   - '{{ mso_site | default("ansible_test") }}'
   - 'aws_{{ mso_site | default("ansible_test") }}'
   - 'azure_{{ mso_site | default("ansible_test") }}'
-  ignore_errors: true
+  register: tenant_site_absent_check
+
+- name: Assertion check for the tenant site dissociation
+  assert:
+    that:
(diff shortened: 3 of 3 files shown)


plugins/modules/mso_tenant_site.py
```diff
@@ -261,6 +261,8 @@ def main():
     # Get tenant_id and site_id
     tenant_id = mso.lookup_tenant(module.params.get("tenant"))
     site_id = mso.lookup_site(module.params.get("site"))
+    if not site_id:
```
akinross, line 264: What would happen on the present/query states? Should we fail or just exit?

plugins/modules/mso_tenant_site.py
```diff
@@ -261,6 +261,8 @@ def main():
     # Get tenant_id and site_id
     tenant_id = mso.lookup_tenant(module.params.get("tenant"))
```
akinross, line 262: Do we have similar behaviour for this lookup method and potentially other lookup methods?

plugins/module_utils/mso.py
```diff
@@ -833,7 +833,8 @@ def lookup_site(self, site):
 
         s = self.get_obj("sites", name=site)
         if not s:
-            self.fail_json(msg="Site '%s' is not a valid site name." % site)
+            self.module.warn("Site '{0}' is not a valid site name.".format(site))
```
akinross, line 836: How do we generically want to handle warnings?

akinross (changes requested):

plugins/modules/mso_tenant_site.py
sajagana: If the site_id is an empty list of None, then the task will simply exit with an empty response(the status will be passed with an empty result).

plugins/modules/mso_tenant_site.py
sajagana: Added ignore_404 parameter to the lookup methods.

plugins/module_utils/mso.py
sajagana: Added ignore_404 parameter to the lookup methods. Based on the ignore_404 value will fail the lookup method.

plugins/module_utils/mso.py
```diff
@@ -776,32 +776,38 @@ def get_obj(self, path, api_version="v1", **kwargs):
             self.fail_json(msg="More than one object matches unique filter: {0}".format(kwargs))
         return objs[0]
 
-    def lookup_schema(self, schema):
+    def lookup_schema(self, schema, ignore_404=False):
```
shrsr, line 779: Few questions ->
1. Do we need to flag ignore_404 as True for all the modules using lookup_schema, lookup_site etc. when the state is absent?
2. If the answer is yes to Q1 can we just modify the functions lookup_schema, lookup_site and etc. without making any change to all the other modules?
Example -
```
def lookup_schema(self, schema):
"""Look up schema and return its id"""
        if schema is None:
            return schema

        schema_summary = self.query_objs("schemas/list-identity", key="schemas", displayName=schema)
        current_state = self.params.get("state")
        if not schema_summary and current_state is not "absent":
            self.fail_json(msg="Provided schema '{0}' does not exist.".format(schema))
        schema_id = schema_summary[0].get("id")
        if not schema_id and current_state is not "absent":
            self.fail_json(msg="Schema lookup failed for schema '{0}': '{1}'".format(schema, schema_id))
        return schema_id
```
3. If we agree on using the flag ignore_404 we could change ignore_404 -> ignore_not_found_error? Let's consult with the others before changing.

shrsr (changes requested):

plugins/module_utils/mso.py
sajagana: Sure @shrsr

akinross (approved):
LGTM

anvitha-jain (approved):
LGTM

lhercot (approved):
LGTM

Issue #377, opened 2023-09-14
Binding type is required in ndo 4.2.1 but not option are available in ansible module


fmabille09:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Binding type is required in ndo 4.2.1 but not option are available in ansible module

### Affected Module Name(s):

cisco.mso.mso_schema_site_anp_epg_domain

### MSO version and MSO Platform

V 4.2.1

### APIC version and APIC Platform for Site Level Resources

V 5.2.7

### Collection versions

cisco.mso 2.2.1

### Output/ Error message

MSO Error 400: site-local EPG XX in site XX in template XX: BindingType is required for VMware VMM Domain

### Expected Behavior

BindingType should be put to default because it's required in ndo now.

### Actual Behavior

BindingType is not available and no default value is put so we get an error message.

### Playbook tasks to Reproduce

Add VMM domain binding to EPG

### References

* #0000

PR #378, opened 2023-09-19
Update mso_schema_site_anp_epg_domain for supporting ndo 4.2 on vmm d…

fmabille09:
…omain

New supported parameters for vmm domain association:
- delimeter
- custom_epg_name
- mac_changes
- forget_transmits
- netflow_pref
- num_port
- port_allocation
- binding_type
- allow_promiscuous

Some of them are mandatory now

### DIFF
--- a/plugins/modules/mso_schema_site_anp_epg_domain.py
+++ b/plugins/modules/mso_schema_site_anp_epg_domain.py
@@ -109,6 +109,47 @@ options:
     description:
     - Distinguished name of EPG lagpolicy. This attribute can only be used with vmmDomain domain association.
     type: str
+  delimeter:
+    description:
+    - Which delimeter to use with this domain association. This attribute can only be used with vmmDomain domain association.
+    type: str
+  binding_type:
+    description:
+    - Which binding_type to use with this domain association. This attribute can only be used with vmmDomain domain association.
+    type: str
+    choices: [ static, dynamic, none, ephemeral ]
+  num_ports:
+    description:
+    - Number of ports for the binding type. This attribute can only be used with vmmDomain domain association.
+    type: str
+  port_allocation:
+    description:
+    - Port allocation for the binding type. This attribute can only be used with vmmDomain domain association and binding type in static.
+    type: str
+  netflow_pref:
+    description:
+    - Which netflow_pref to use with this domain association. This attribute can only be used with vmmDomain domain association.
+    type: str
+    choices: [ enabled, disabled ]
+  allow_promiscuous:
+    description:
+    - Which allow_promiscuous to use with this domain association. This attribute can only be used with vmmDomain domain association.
+    type: str
+    choices: [ accept, reject ]
+  forged_transmits:
+    description:
+    - Which forged_transmits to use with this domain association. This attribute can only be used with vmmDomain domain association.
+    type: str
+    choices: [ accept, reject ]
+  mac_changes:
+    description:
+    - Which mac_changes to use with this domain association. This attribute can only be used with vmmDomain domain association.
+    type: str
+    choices: [ accept, reject ]
+  custom_epg_name:
+    description:
+    - Which custom_epg_name to use with this domain association. This attribute can only be used with vmmDomain domain association.
+    type: str
   state:
     description:
     - Use C(present) or C(absent) for adding or removing.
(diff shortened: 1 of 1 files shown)


fmabille09:
In response to issues #377

plugins/modules/mso_schema_site_anp_epg_domain.py
```diff
@@ -222,14 +263,23 @@ def main():
         switching_mode=dict(type="str"),
         enhanced_lagpolicy_name=dict(type="str"),
         enhanced_lagpolicy_dn=dict(type="str"),
+        binding_type=dict(type="str", default="none", choices=["dynamic", "ephemeral", "none", "static"]),
```
akinross, line 266: I am missing the following items:
- test updates in https://github.com/CiscoDevNet/ansible-mso/blob/master/tests/integration/targets/mso_schema_site_anp_epg_domain/tasks/main.yml
- example update with the new attributes

Could you please add this?

plugins/modules/mso_schema_site_anp_epg_domain.py
```diff
@@ -392,8 +451,14 @@ def main():
...
         deployImmediacy=deployment_immediacy,  # rename of deploymentImmediacy
         resolutionImmediacy=resolution_immediacy,
+        bindingType=binding_type,
+        netflowPref=netflow_pref,
+        allowPromiscuous=allow_promiscuous,
+        forgedTransmits=forged_transmits,
+        macChanges=mac_changes
     )
 
+
```
akinross, line 461: not sure why additional white line is added, can be removed?

plugins/modules/mso_schema_site_anp_epg_domain.py
```diff
@@ -414,11 +479,27 @@ def main():
 
         if vlan_encap_mode:
             vmmDomainProperties["vlanEncapMode"] = vlan_encap_mode
+
```
akinross, line 482: not sure why additional white line is added, can be removed?

plugins/modules/mso_schema_site_anp_epg_domain.py
akinross: Also if possible please add your information to the author section and the top of the file

plugins/modules/mso_schema_site_anp_epg_domain.py
```diff
@@ -222,14 +263,23 @@ def main():
...
+        custom_epg_name=dict(type="str"),
     )
 
     module = AnsibleModule(
         argument_spec=argument_spec,
         supports_check_mode=True,
         required_if=[
             ["state", "absent", ["domain_association_type", "domain_profile", "deployment_immediacy", "resolution_immediacy"]],
-            ["state", "present", ["domain_association_type", "domain_profile", "deployment_immediacy", "resolution_immediacy"]],
+            ["state", "present", ["domain_association_type", "domain_profile", "deployment_immediacy", "resolution_immediacy", "binding_type", "netflow_pref", "allow_promiscuous", "forged_transmits", "mac_changes"]],
```
akinross, line 282: Making these new attributes required on present is version depending and will likely break existing functionality. So don't think these should be made required on present. If we want to catch this we would need to do a version check. What is the error response from NDO when it is not provided in 4.2.x?

plugins/modules/mso_schema_site_anp_epg_domain.py
```diff
@@ -109,6 +109,47 @@
     description:
     - Distinguished name of EPG lagpolicy. This attribute can only be used with vmmDomain domain association.
     type: str
+  delimeter:
+    description:
+    - Which delimeter to use with this domain association. This attribute can only be used with vmmDomain domain association.
+    type: str
+  binding_type:
```
akinross, line 116: I think we should add to the description that this is required in the version where it got introduced. Same applies if there are such constraints for the other attributes.

akinross (changes requested):

plugins/modules/mso_schema_site_anp_epg_domain.py
fmabille09: I removed it in the old commit

plugins/modules/mso_schema_site_anp_epg_domain.py
fmabille09: I removed it in commit

plugins/modules/mso_schema_site_anp_epg_domain.py
fmabille09: Yes I think we shouldn't put it in required to don't break other versions. 
The error message is the following: MSO Error 400: site-local EPG XX in site XX in template XX: BindingType is required for VMware VMM Domain and is the same for other attributes.

plugins/modules/mso_schema_site_anp_epg_domain.py
akinross: For me it is about informing the user that it is a required attribute when on a specific version (4.2 I think but this would need to be verified) . The PR is fixing the issue: MSO Error 400: site-local EPG XX in site XX in template XX: BindingType is required for VMware VMM Domain. So I would say that a additional description would be good to have, since we are not enforcing in the required_if. @lhercot, @shrsr, @anvitha-jain, @sajagana what is your opinion on this?

plugins/modules/mso_schema_site_anp_epg_domain.py
akinross: @fmabille09 was the testing part clear, do you need any assistance on this?

akinross:
@fmabille09 was the testing part clear, do you need any assistance on this?

akinross:
PR not finished by raiser, see new PR https://github.com/CiscoDevNet/ansible-mso/pull/403

Issue #383, opened 2023-10-24
Unable to set schema and template for L3out in mso_schema_site_external_epg


mnejedlo-tds:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

* While attempting to use mso_schema_site_external_epg, I ran into a problem where the l3out parameter assumes the same schema and template name as for the eEPG, but having the eEPG and L3out in different schemas/templates is an allowed configuration in NDO.  Please update mso_schema_site_external_epg to use a name/schema/template dictionary for the l3out parameter, similar to how mso_schema_template_external_epg works currently.

### New or Affected Module(s):

* mso_schema_site_external_epg

### MSO version and MSO Platform

* V 3.6.1e on ND 2.1(2d) (VMware/OVA based)

### APIC version and APIC Platform for Site Level Resources

* V 5.2(3g) on-prem

### Collection versions

* cisco.mso 2.5.0

shrsr:
@mnejedlo-tds Thank you for opening the issue. Our team will look into it and come up with a solution.

PR #394, opened 2023-12-18
[minor_changes] Added L3Out Argument at site external EPG module

anvitha-jain:
solves #383 

Associating L3Out created in different templates is only possible in site external EPGs created in NDO version < 4.0.

### DIFF
--- a/plugins/modules/mso_schema_site_external_epg.py
+++ b/plugins/modules/mso_schema_site_external_epg.py
@@ -13,67 +13,61 @@ ANSIBLE_METADATA = {"metadata_version": "1.1", "status": ["preview"], "supported
 DOCUMENTATION = r"""
 ---
 module: mso_schema_site_external_epg
-short_description: Manage External EPG in schema of sites
+short_description: Manage External EPG in schema of sites.
 description:
 - Manage External EPG in schema of sites on Cisco ACI Multi-Site.
-- This module can only be used on versions of MSO that are 3.3 or greater
+- This module can only be used on versions of MSO that are 3.3 or greater.
 author:
 - Anvitha Jain (@anvitha-jain)
 options:
   schema:
     description:
-    - The name of the schema
+    - The name of the schema.
     type: str
     required: true
   template:
     description:
-    - The name of the template to change
+    - The name of the template to change.
     type: str
     required: true
   l3out:
     description:
-    - The L3Out associated with the external epg
-    - Required when site is of type on-premise
+    - The L3Out associated with the external epg.
+    - Required when site is of type on-premise.
     - NOTE - for NDO version > 4.2, this parameter is available only when external EPG is 
-    - associated with the VRF created at the current schema-template
-    type: dict
-    suboptions:
-      name:
-        description:
-        - The name of the L3Out to associate with
-        required: true
-        type: str
-      schema:
-        description:
-        - The schema that defines the referenced L3Out
-        - If this parameter is unspecified, it defaults to the current schema
-        type: str
-      template:
-        description:
-        - The template that defines the referenced L3Out
-        - If this parameter is unspecified, it defaults to the current template
-        type: str
+    - associated with the VRF created at the current schema-template.
+    type: str
+  l3out_schema:
+    description:
+    - The schema that defines the referenced L3Out.
+    - If this parameter is unspecified, it defaults to the current schema.
+    type: str
+  l3out_template:
+    description:
+    - The template that defines the referenced L3Out.
+    - If this parameter is unspecified, it defaults to the current template.
+    type: str
   external_epg:
     description:
-    - The name of the External EPG to be managed
+    - The name of the External EPG to be managed.
     type: str
     aliases: [ name ]
   site:
     description:
-    - The name of the site
+    - The name of the site.
     type: str
     required: true
   route_reachability:
     description:
-    - Configures if an external EPG route is pointing to the internet or to an external remote network
-    - Only available when associated with an azure site
+    - Configures if an external EPG route is pointing to the internet or to an external remote network.
+    - Only available when associated with an azure site.
     type: str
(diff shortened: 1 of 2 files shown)


plugins/modules/mso_schema_site_external_epg.py
```diff
@@ -16,46 +16,64 @@
...
+    - The name of the schema
     type: str
     required: true
   template:
     description:
-    - The name of the template to change.
+    - The name of the template to change
     type: str
     required: true
   l3out:
```
akinross, line 33: Should we introduce as separate input strings instead of dictionary in order to avoid breaking change? So l3out ( with alias l3out_name), l3out_schema, l3out_template as string inputs.

plugins/modules/mso_schema_site_external_epg.py
```diff
@@ -16,46 +16,64 @@
 short_description: Manage External EPG in schema of sites
 description:
 - Manage External EPG in schema of sites on Cisco ACI Multi-Site.
-- This module can only be used on versions of MSO that are 3.3 or greater.
+- This module can only be used on versions of MSO that are 3.3 or greater
```
akinross, line 19: Why are we removing all the dots behind sentences?

akinross (changes requested):

plugins/modules/mso_schema_site_external_epg.py
anvitha-jain: Added back as discussed

plugins/modules/mso_schema_site_external_epg.py
anvitha-jain: not removing as discussed

plugins/modules/mso_schema_site_external_epg.py
```diff
@@ -73,7 +85,8 @@
     schema: Schema 1
     template: Template 1
     external_epg: External EPG 1
-    l3out: L3out1
+    l3out:
```
akinross, line 88: this should not be dict input anymore correct?

plugins/modules/mso_schema_site_external_epg.py
```diff
@@ -84,7 +97,8 @@
...
```
akinross, line 100: this should not be dict input anymore correct?

akinross (changes requested):

plugins/modules/mso_schema_site_external_epg.py
```diff
@@ -34,6 +34,18 @@
     description:
     - The L3Out associated with the external epg.
     - Required when site is of type on-premise.
+    - NOTE - for NDO version > 4.2, this parameter is available only when external EPG is
```
shrsr, line 37: Can you please check if the below implies the same meaning?

- In NDO versions over 4.2, the parameter is accessible only with an external EPG linked to the current schema-template's VRF.

shrsr (changes requested):

shrsr:
Can you also rebase?

plugins/modules/mso_schema_site_external_epg.py
anvitha-jain: yes and changed

plugins/modules/mso_schema_site_external_epg.py
anvitha-jain: yes changed

plugins/modules/mso_schema_site_external_epg.py
anvitha-jain: yes changed

plugins/modules/mso_schema_site_external_epg.py
```diff
@@ -144,11 +158,17 @@ def main():
...
     external_epg = module.params.get("external_epg")
     l3out = module.params.get("l3out")
+    l3out_schema = module.params.get("l3out_schema")
+    l3out_template = module.params.get("l3out_template")
     route_reachability = module.params.get("route_reachability")
     state = module.params.get("state")
 
     mso = MSOModule(module)
 
+    l3out_template = template if l3out_template is None else l3out_template.replace(" ", "")
```
samiib, line 168: Why are we specifically removing spaces from the template name? Unless I am missing anything.

samiib: Ah okay, good to know thanks.

samiib (approved):
LGTM

plugins/modules/mso_schema_site_external_epg.py
samiib: Then just a suggestion would be to document that behaviour in the description of l3out_template.

plugins/modules/mso_schema_site_external_epg.py
```diff
@@ -34,6 +34,18 @@
     description:
     - The L3Out associated with the external epg.
     - Required when site is of type on-premise.
+    - NOTE - In NDO versions over 4.2,
```
shrsr, line 37: Can we remove the 'NOTE -' and just include the point as is?

I think the rest looks good.

shrsr (changes requested):

plugins/modules/mso_schema_site_external_epg.py
anvitha-jain: done

plugins/modules/mso_schema_site_external_epg.py
akinross, line 168: Why do we only do a replace() function on l3out_template and not on template variable assignment?

akinross (changes requested):

akinross (approved):
LGTM

shrsr (approved):
LGTM

samiib (approved):
LGTM

lhercot (approved):
LGTM

PR #404, opened 2024-02-06
Added Azure Cloud CNC support to the mso_schema_site_service_graph module

sajagana:
Note:

Added Azure Cloud CNC support to the mso_schema_site_service_graph module and fixed mso_schema_template_service_graph module.

### DIFF
--- a/plugins/module_utils/mso.py
+++ b/plugins/module_utils/mso.py
@@ -859,11 +859,11 @@ class MSOModule(object):
             ids.append(dict(roleId=r.get("id"), accessType=access_type))
         return ids
 
-    def lookup_site_type(self, s):
+    def lookup_site_type(self, site_data):
         """Get site type(AWS, AZURE or physical)"""
-        site_type = s.get("platform")
+        site_type = site_data.get("platform")
         if site_type == "cloud":
-            self.cloud_provider_type = s.get("cloudProviders")[0]
+            self.cloud_provider_type = site_data.get("cloudProviders")[0]
         self.site_type = site_type
 
     def lookup_site(self, site, ignore_not_found_error=False):
--- a/plugins/module_utils/mso.py
+++ b/plugins/module_utils/mso.py
@@ -22,7 +22,7 @@ from ansible.module_utils.six.moves.urllib.parse import urlencode, urljoin
 from ansible.module_utils.urls import fetch_url
 from ansible.module_utils._text import to_native, to_text
 from ansible.module_utils.connection import Connection
-from ansible_collections.cisco.mso.plugins.module_utils.constants import NDO_API_VERSION_PATH_FORMAT
+from ansible_collections.cisco.mso.plugins.module_utils.constants import NDO_API_VERSION_PATH_FORMAT, AZURE_L4L7_CONNECTOR_TYPE_MAP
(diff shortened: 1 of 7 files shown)


samiib (changes requested):
There are some long lines sanity errors.

tests/integration/targets/mso_schema_site_service_graph/tasks/l4_l7_devices.yml
```diff
@@ -0,0 +1,263 @@
...
+              - cloudRsCtxProfileToRegion:
+                  attributes:
+                    tDn: "uni/clouddomp/provp-azure/region-westus2"
+              - cloudCidr:
+                  attributes:
+                    addr: 1.1.1.2/32
+                    primary: "yes"
+
+    - name: Add the subnet to the sab_azure_ctxprof
+      cisco.aci.aci_rest:
```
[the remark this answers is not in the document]
sajagana: aci_cloud_subnet - has some issues, so using aci_rest to manage it.

samiib (approved):
LGTM! Yes will try and fix the aci_cloud_subnet shortly. :)

plugins/module_utils/mso.py
```diff
@@ -853,6 +859,13 @@ def lookup_roles(self, roles, ignore_not_found_error=False):
             ids.append(dict(roleId=r.get("id"), accessType=access_type))
         return ids
 
+    def lookup_site_type(self, s):
```
akinross, line 862: Can you rename "s" to a more descriptive value. Found it confusing to understand where this platform key is now coming from. This is also due to the fact that in the module class we also have a platform defined with values "local", "nd" and "mso".

plugins/modules/mso_schema_site_service_graph.py
```diff
@@ -225,29 +280,77 @@ def main():
...
-                        serviceNodeRef=dict(
-                            serviceNodeName=template_node_type,
-                            serviceGraphName=service_graph,
-                            templateName=template,
-                            schemaId=schema_id,
-                        ),
-                    ),
-                )
+                device_payload = dict()
+                if query_device_data:
```
akinross, line 286: So when query_device_data is None and empty dict will still be added to the payload? Is there a reason for this?

plugins/modules/mso_schema_site_service_graph.py
```diff
@@ -225,29 +280,77 @@ def main():
...
+                    for device_data in query_device_data:
+                        if device.get("device_name") == device_data.get("dn").split("/")[-1].split("-")[-1]:
```
akinross, line 288: Can `device_data.get("dn").split("/")[-1].split("-")[-1]` ever lead to an error?

sajagana: The DN value points to the Cloud Site L4-L7 device, so it should follow the same pattern. Also, the DN format is common for the current L4-L7 devices.

akinross: ok

akinross (approved):
LGTM

gmicol (approved):
LGTM

lhercot (approved):
LGTM

PR #430, opened 2024-06-12
[minor_change] Added module ndo_template to support creation of tenan…

akinross:
…t, l3out, fabric_policy, fabric_resource, monitoring_tenant, monitoring_access and service_device templates

### DIFF
--- a/plugins/module_utils/template.py
+++ b/plugins/module_utils/template.py
@@ -1,0 +1,88 @@
+# -*- coding: utf-8 -*-
+
+# Copyright: (c) 2022, Akini Ross (@akinross) <akinross@cisco.com>
+# GNU General Public License v3.0+ (see LICENSE or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+
+__metaclass__ = type
+
+from ansible_collections.cisco.mso.plugins.module_utils.constants import TEMPLATE_TYPES
+from collections import namedtuple
+
+KVPair = namedtuple("KVPair", "key value")
+Item = namedtuple("Item", "index details")
+SearchQuery = namedtuple("SearchQuery", "key kv_pairs")
+
+
+class MSOTemplate:
+    def __init__(self, mso_module, template_type=None, template_name=None, template_id=None):
+        self.mso = mso_module
+        self.templates_path = "templates"
+        self.summaries_path = "{0}/summaries".format(self.templates_path)
+        self.template = {}
+        self.template_path = ""
+        self.template_name = template_name
+        self.template_id = template_id
+
+        if template_id:
+            # Checking if the template with id exists to avoid error: MSO Error 400: Template ID 665da24b95400f375928f195 invalid
+            template_summary = self.mso.get_obj(self.summaries_path, templateId=self.template_id)
+            if template_summary:
+                self.template_path = "{0}/{1}".format(self.templates_path, self.template_id)
+                self.template = self.mso.query_obj(self.template_path)
+            else:
+                self.mso.fail_json(
+                    msg="Provided template id '{0}' does not exist. Existing templates: {1}".format(
+                        self.template_id,
+                        [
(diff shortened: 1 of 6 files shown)


akinross:
requires rebase after https://github.com/CiscoDevNet/ansible-mso/pull/428 has been merged

anvitha-jain (approved):
LGTM

sajagana (dismissed):
LGTM!

plugins/module_utils/template.py
```diff
@@ -0,0 +1,88 @@
...
+        :param search_list: Objects to search through -> List.
+        :param kv_list: Key/value pairs that should match in the object. -> List[KVPair(Str, Str)]
+        :return: The index and details of the object. -> Item (Named Tuple)
+                 Values of provided keys of all existing objects. -> List
+        """
+
+        def kv_match(kvs, item):
+            return all((item.get(kv.key) == kv.value for kv in kvs))
+
+        match = next((Item(index, item) for index, item in enumerate(search_list) if kv_match(kv_list, item)), None)
```
samiib, line 76: I am wondering about the performance of this method. 
Would it be possible to have a very large `search_list` and `kv_list`?
If not then performance may not really be an issue.

samiib: I can see from the other PR's that `kv_list` is mostly just 1 item like 'uuid' or 'name'.
But could `search_list` contain a large number of objects?

akinross: There are occasions where it might be required, like we had on schema for schema id and template name for instance. It is most likely to be one item, but this provides the flexibility to search for more without adding new functionality. The problem is here that we need to match on values inside of lists instead of keys in dicts, due to how the response of schema/template APIs. If it is something you think is valuable we could have a look into performance analysis for this?

The search list size, depends on the configuration of customer. I do not know the exact size of this, but my assumption is that is the amount of objects will not even be thousands. Furthermore I think the bottle-neck with anything we do is not search operations but the I/O operations towards NDO.

shrsr: Where is this function used?

shrsr (changes requested):

plugins/module_utils/template.py
akinross: If you refer to kv_match() that is used in line 76: `match = next((Item(index, item) for index, item in enumerate(search_list) if kv_match(kv_list, item)), None)`

shrsr: Sorry! I meant get_object_from_list(). I haven't checked the other PRs so maybe it is used there?

shrsr (dismissed):
LGTM

plugins/module_utils/template.py
akinross: Indeed see the following PRs:

https://github.com/CiscoDevNet/ansible-mso/pull/472/files#diff-3409e22b7188780872a6a0c90a47c28744a04759f0f5d53c5e5dfd5587fed0b4

https://github.com/CiscoDevNet/ansible-mso/pull/471/files#diff-3409e22b7188780872a6a0c90a47c28744a04759f0f5d53c5e5dfd5587fed0b4

https://github.com/CiscoDevNet/ansible-mso/pull/435/files#diff-3409e22b7188780872a6a0c90a47c28744a04759f0f5d53c5e5dfd5587fed0b4

https://github.com/CiscoDevNet/ansible-mso/pull/434/files#diff-3409e22b7188780872a6a0c90a47c28744a04759f0f5d53c5e5dfd5587fed0b4

samiib: Makes sense, if search list is within the 100's I don't think we will see much of a performance impact with this code specifically. Maybe just something to take note of to look into if anyone does report performance issues.

samiib (approved):
LGTM

shrsr (approved):
LGTM

samiib (approved):
LGTM

lhercot (approved):
LGTM

Issue #426, opened 2024-05-29
Add BD route-map source and destination filter capablity


sauguieb:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description
Since release APIC 5.0.1 and in NDO 3.7.2, multicast filtering is available on NDO (router-map source/destination filter).

### New or Affected Module(s):

* cisco.mso.mso_schema_template_bd module 

### MSO version and MSO Platform

* V 3.7.2 and all.

### APIC version and APIC Platform for Site Level Resources

* V 5.2. and all.

### Collection versions

* all

### References

N/A

* #0000

Issue #425, opened 2024-05-29
Add multicast RP configuration on VRF


sauguieb:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description
In release MSO 3.0(1) and APIC 5.0(1), multicast RP can reside in the fabric.
We would like to be able to configure it from NDO which includes
- RP ip address
- type: Fabric or static
- route-map poliy

### New or Affected Module(s):

cisco.mso.mso_schema_template_vrf module

### MSO version and MSO Platform

* V 4.2.3 and all.

### APIC version and APIC Platform for Site Level Resources

* V 5.2 and all.

### Collection versions

all

### References

N/A

* #0000

PR #435, opened 2024-06-14
Tenant policy multicast route map policy

akinross:
needed for #426 and https://github.com/CiscoDevNet/ansible-mso/issues/425

### DIFF
--- a/plugins/modules/ndo_route_map_policy_multicast.py
+++ b/plugins/modules/ndo_route_map_policy_multicast.py
@@ -1,0 +1,269 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# Copyright: (c) 2024, Akini Ross (@akinross) <akinross@cisco.com>
+
+# GNU General Public License v3.0+ (see LICENSE or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+
+__metaclass__ = type
+
+ANSIBLE_METADATA = {"metadata_version": "1.1", "status": ["preview"], "supported_by": "community"}
+
+DOCUMENTATION = r"""
+---
+module: ndo_route_map_policy_multicast
+short_description: Manage Multicast Route Map Policies on Cisco Nexus Dashboard Orchestrator (NDO).
+description:
+- Manage Multicast Route Map Policies on Cisco Nexus Dashboard Orchestrator (NDO).
+author:
+- Akini Ross (@akinross)
+options:
+  template:
+    description:
+    - The name of the template.
+    - The template must be a tenant template.
+    type: str
+    required: true
+  route_map_policy:
+    description:
+    - The name of the Multicast Route Map Policy.
+    type: str
+    aliases: [ name ]
+  route_map_policy_uuid:
+    description:
+    - The uuid of the Multicast Route Map Policy.
+    - This parameter is required when the O(route_map_policy) needs to be updated.
+    type: str
+    aliases: [ uuid ]
+  description:
+    description:
+    - The description of the Multicast Route Map Policy.
+    type: str
+  entries:
+    description:
+    - A list of entries attached to the Multicast Route Map Policy.
+    - The list of configured entries must contain at least one entry.
+    - When the list of entries is null the update will not change existing entry configuration.
+    type: list
+    elements: dict
+    suboptions:
+      order:
+        description:
+        - The order of the entry.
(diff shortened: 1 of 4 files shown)


akinross:
requires rebase after https://github.com/CiscoDevNet/ansible-mso/pull/430 is merged

sajagana (approved):
LGTM!

tests/integration/targets/ndo_template/tasks/main.yml
```diff
@@ -0,0 +1,592 @@
...
+
+    - name: Error on not allowed tenant
+      cisco.mso.ndo_template:
+        <<: *create_monitoring_access_template
+        tenant: ansible_test
+        state: present
+      register: err_not_allowed_tenant
+      ignore_errors: true
+
+    - name: Err on template query without template_type
```
anvitha-jain, line 490: Err -> Error

anvitha-jain (changes requested):

tests/integration/targets/ndo_template/tasks/main.yml
akinross: Changed and rebased all the PRs that are affected by this

sajagana (approved):
LGTM!

anvitha-jain (approved):
LGTM

sajagana (approved):
LGTM!

plugins/modules/ndo_route_map_policy_multicast.py
```diff
@@ -0,0 +1,269 @@
...
+      source:
+        description:
+        - The data of the entry.
+        type: str
+        required: true
+      rp:
+        description:
+        - The rendezvous point of the entry.
+        type: str
+        aliases: [ rendezvous_point ]
```
samiib, line 71: I think the name and the alias should be switched.
```suggestion
      rendezvous_point:
        description:
        - The rendezvous point of the entry.
        type: str
        aliases: [ rp ]
```

plugins/modules/ndo_route_map_policy_multicast.py
```diff
@@ -0,0 +1,269 @@
...
+    - The name of the Multicast Route Map Policy.
+    type: str
+    aliases: [ name ]
+  route_map_policy_uuid:
+    description:
+    - The uuid of the Multicast Route Map Policy.
+    - This parameter is required when the O(route_map_policy) needs to be updated.
+    type: str
+    aliases: [ uuid ]
+  description:
```
samiib, line 40: Should we add some aliases for description?

samiib (changes requested):

plugins/modules/ndo_route_map_policy_multicast.py
akinross: why?

plugins/modules/ndo_route_map_policy_multicast.py
akinross: For multiple other modules ( see template modules where description ) we do not add an alias, so do not see why we should now introduce here. If so perhaps raise an issue and change it consistently everywhere.

anvitha-jain (approved):
LGTM

plugins/modules/ndo_route_map_policy_multicast.py
samiib: As a general option convention, If I was adding the full name as an option/alias, I would use shorter names and alternatives in alias. However, it is up to you as there is no discernible benefit either way.

plugins/modules/ndo_route_map_policy_multicast.py
samiib: Its fine to keep it consistent. I was just thinking about what we usually do in `ansible-aci` for description.

plugins/modules/ndo_route_map_policy_multicast.py
akinross: ok will keep as is

plugins/modules/ndo_route_map_policy_multicast.py
akinross: Yeah I know there we use `descr`, but I think this is also due to the model in apic using `descr`. Feel free to open an new issue for changing everywhere if you feel this is an improvement.

samiib (approved):
LGTM

gmicol (approved):
LGTM

lhercot (approved):
LGTM

PR #564, opened 2024-11-08
[ignore] Added codeowners file so the DCN Automation team can be auto added as PR reviewers. (DCNE-230)

samiib:


lhercot (approved):
LGTM

PR #563, opened 2024-11-08
Added ndo_l3out_interface_routing_policy to manage L3Out Interface Routing Policy (DCNE-132)

sajagana:

PR #569, opened 2024-11-15
Added append_update_ops_data to handle the latest ndo module update operations (DCNE-245)

sajagana:
Note:

Used the same function to test the code with this PR: https://github.com/CiscoDevNet/ansible-mso/pull/563

### DIFF
--- a/plugins/module_utils/utils.py
+++ b/plugins/module_utils/utils.py
@@ -30,50 +30,101 @@ def generate_api_endpoint(path, **kwargs):
     return path if not kwargs else "{0}?{1}".format(path, "&".join(["{0}={1}".format(key, value) for key, value in kwargs.items()]))
 
 
-def append_update_ops_data(ops, existing_data, update_path, value, *args, op="replace"):
+def append_update_ops_data(ops, existing_data, update_path, value, *existing_data_keys, op="replace"):
     """
     Append Update ops payload data.
     :param ops: Variable which contains the PATCH replace actions for the update operation -> List
     :param existing_data: Variable which contains the existing data -> Dict
     :param update_path: The object path is used to update an existing object -> Str
     :param value: Input value of the attribute which needs to be updated -> Any
-    :param args: Extra arguments which is used to compare and update the existing data -> Any
+    :param existing_data_keys: Additional positional arguments which is used to compare and update the existing data based on the sequence of keys -> Any
     :param op: Defaults to "replace" when not specified, value "remove" is used clear the existing configuration -> Str
     :return: None
-                If args is not empty then the ops and existing_data are updated with the input value.
+                If attributes is not empty then the ops and existing_data are updated with the input value.
+
+    Sample Inputs:
+        ops = []
+        existing_data = {
+            "name": "L3OutInterfacePolicy",
+            "bfdMultiHopPol": {
+                "adminState": "enabled",
+                "detectionMultiplier": 3,
+                "minRxInterval": 250,
+                "minTxInterval": 250
+            },
+        }
+        update_path = "/tenantPolicyTemplate/template/l3OutIntfPolGroups/0"
+
+        Sample existing_data_keys input format:
+            To update the "bfdMultiHopPol -> adminState" value:
+                existing_data_keys = "bfdMultiHopPol", "adminState"
+
+        To update the top level key of the existing object:
+            append_update_ops_data(ops, existing_data, update_path, "L3OutInterfacePolicy_NameUpdated", "name", op="replace")
+
+        To create the nested object from the scratch when the object is not available:
+            append_update_ops_data(ops, existing_data, update_path, dict(), "ospfIntfPol", op="replace")
+            append_update_ops_data(ops, existing_data, update_path, dict(), "ospfIntfPol", "ifControl", op="replace")
+            append_update_ops_data(ops, existing_data, update_path, True, "ospfIntfPol", "ifControl", "ignoreMtu", op="replace")
+
+        After inserting the new object:
+        existing_data = {
+            "name": "L3OutInterfacePolicy",
+            "ospfIntfPol": {
+                "networkType": "broadcast",
+                "prio": 1,
+                "cost": 0,
+                "ifControl": {
+                    "advertiseSubnet": false,
+                    "bfd": false,
+                    "ignoreMtu": false,
+                    "passiveParticipation": false
+                },
+                "helloInterval": 10,
+                "deadInterval": 40,
+                "retransmitInterval": 5,
+                "transmitDelay": 1
+            },
+        }
+
+        To update the nested key of the existing object:
+            append_update_ops_data(ops, existing_data, update_path, True, "ospfIntfPol", "ifControl", "ignoreMtu", op="replace")
+
+        To delete the nested object from the existing object:
+            append_update_ops_data(ops, existing_data, update_path, None, "ospfIntfPol", "ifControl", op="remove")
+
+        To delete the top level key from the existing object:
+            append_update_ops_data(ops, existing_data, update_path, None, "ospfIntfPol", op="remove")
     """
 
-    def recursive_update(data, current_path, keys, new_value):
+    def recursive_update(data, path, keys, new_value):
         if len(keys) == 0:
             return
         key = keys[0]
 
         if len(keys) == 1:
             # Update the existing configuration
-            if new_value is not None and data.get(key) != new_value:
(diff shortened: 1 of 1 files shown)


plugins/module_utils/utils.py
```diff
@@ -26,3 +28,52 @@ def generate_api_endpoint(path, **kwargs):
...
+                data[key] = new_value
+                ops.append(
+                    dict(
+                        op=op,
+                        path="{}/{}".format(current_path, key),
+                        value=copy.deepcopy(new_value),
+                    )
+                )
+            # Clear the existing configuration
+            elif op == "remove":
```
akinross, line 63: do we always add op on remove or should we check data that the "object" to remove actually exist in data?

plugins/module_utils/utils.py
```diff
@@ -26,3 +28,52 @@ def generate_api_endpoint(path, **kwargs):
...
+                data.pop(key, None)
+                ops.append(
+                    dict(
+                        op=op,
+                        path="{}/{}".format(current_path, key),
+                    )
+                )
+
+        elif key in data:
```
akinross, line 72: Should we handle the else clause where the key is not in data?

plugins/module_utils/utils.py
```diff
@@ -26,3 +28,52 @@ def generate_api_endpoint(path, **kwargs):
...
+            recursive_update(data[key], "{}/{}".format(current_path, key), keys[1:], new_value)
+
+    if args:
+        if isinstance(args[0], tuple):
```
akinross, line 76: when would args[0] be a tuple?

plugins/module_utils/utils.py
```diff
@@ -26,3 +28,52 @@ def generate_api_endpoint(path, **kwargs):
...
+    :param existing_data: Variable which contains the existing data -> Dict
+    :param update_path: The object path is used to update an existing object -> Str
+    :param value: Input value of the attribute which needs to be updated -> Any
+    :param args: Extra arguments which is used to compare and update the existing data -> Any
+    :param op: Defaults to "replace" when not specified, value "remove" is used clear the existing configuration -> Str
+    :return: None
+                If args is not empty then the ops and existing_data are updated with the input value.
+    """
+
+    def recursive_update(data, current_path, keys, new_value):
```
akinross, line 46: should we shorten current_path to path?

plugins/module_utils/utils.py
```diff
@@ -26,3 +28,52 @@ def generate_api_endpoint(path, **kwargs):
     # return full_url
 
     return path if not kwargs else "{0}?{1}".format(path, "&".join(["{0}={1}".format(key, value) for key, value in kwargs.items()]))
+
+
+def append_update_ops_data(ops, existing_data, update_path, value, *args, op="replace"):
+    """
+    Append Update ops payload data.
```
akinross, line 35: Can we elaborate the function bit more? Perhaps add a call example?

plugins/module_utils/utils.py
sajagana: If the key is new and a leaf key of the object, that will be handled in the if part.

When the `len(keys) == 1`, then that key will be added to the payload even if the key does not exist in the data.

samiib (changes requested):

plugins/module_utils/utils.py
```diff
@@ -26,3 +28,103 @@ def generate_api_endpoint(path, **kwargs):
...
+                data[key] = new_value
+                ops.append(
+                    dict(
+                        op=op,
+                        path="{}/{}".format(path, key),
+                        value=copy.deepcopy(new_value),
+                    )
+                )
+            # Clear the existing configuration
+            elif op == "remove" and key in data:
```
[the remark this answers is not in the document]
sajagana: We can throw error when the input is empty and `op` is not `replace` or `remove`.
Can we return the error message or fail the module execution?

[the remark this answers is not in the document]
samiib: Maybe it should fail the module. It will most likely be an issue when modules are first developed and testing would quickly pick up the failure.
Does throwing an exception get caught as a module failure in Ansible?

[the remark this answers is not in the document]
gmicol: @samiib It does get caught and the module will therefore fail. I think we should throw an error as well just in case:
```Python
valid_ops = ["replace", "remove"]
if op not in valid_ops:
  raise ValueError("Invalid op value. Expected one of: {0}. Got: {1}".format(valid_ops, op))
```

plugins/module_utils/utils.py
```diff
@@ -26,3 +28,173 @@ def generate_api_endpoint(path, **kwargs):
...
+        "name": "irp_1",
+    }
+    ops = []
+    update_path = "/tenantPolicyTemplate/template/l3OutIntfPolGroups/0"
+
+    replace_data = {
+        ("name"): "new_name",
+        "description": "new_description",
+        ("bfdMultiHopPol", "ifControl"): dict(),
+        ("bfdMultiHopPol", "ifControl", "adminState"): "enabled",
```
gmicol, line 61: Is there a scenario where it's interesting to put an empty dict? otherwise, we should merge these two examples into one:
```suggestion
        ("bfdMultiHopPol", "ifControl"): dict(adminState="enabled"),
```

gmicol (changes requested):
Apart from throwing a few extra errors in specific cases, the overall structure of the function looks good to me

plugins/module_utils/utils.py
sajagana: The current code supports merging the two lines into single-line logic.

shrsr (dismissed):
LGTM

samiib (approved):
LGTM

plugins/module_utils/utils.py
```diff
@@ -26,3 +28,174 @@ def generate_api_endpoint(path, **kwargs):
...
+            "ifControl": {
+                "adminState": "disbaled",
+            },
+            "cost": 0,
+        },
+    }
+    ops = []
+    update_path = "/tenantPolicyTemplate/template/l3OutIntfPolGroups/0"
+
+    remove_data = {
```
akinross, line 133: Do we need this to be a dictionary, because we only require the keys for removal?

```suggestion
    remove_data = [key1, key2]
```

plugins/module_utils/utils.py
```diff
@@ -26,3 +28,174 @@ def generate_api_endpoint(path, **kwargs):
     # return full_url
 
     return path if not kwargs else "{0}?{1}".format(path, "&".join(["{0}={1}".format(key, value) for key, value in kwargs.items()]))
+
+
+def append_update_ops_data(ops, existing_data, update_path, changes=None, op="replace"):
```
akinross, line 33: Should we expose replace and remove as two options, this way we do not have to specify the operation which is deduced from the passed in attribute. Furthermore this way we could process all operation in a single function call.

```suggestion
def append_update_ops_data(ops, existing_data, update_path, replace_data=None, remove_list=None):

if replace_data is None:
  replace_data = {}

if remove_data is None:
  remove_list = []
```

akinross (changes requested):

akinross (dismissed):
LGTM

samiib (approved):
LGTM

gmicol (dismissed):
LGTM

plugins/module_utils/utils.py
```diff
@@ -26,3 +28,124 @@ def generate_api_endpoint(path, **kwargs):
...
+                ops.append(
+                    dict(
+                        op="remove",
+                        path="{}/{}".format(path, key),
+                    )
+                )
+        elif key in data:
+            recursive_delete(data[key], "{}/{}".format(path, key), keys[1:])
+
+    if isinstance(replace_data, dict):
```
shrsr, line 135: Shall we do this instead?
Same for remove data?
```
if not isinstance(replace_data, dict):
    raise TypeError("replace_data must be a dict")

for key, value in replace_data.items():
    keys_tuple = key if isinstance(key, tuple) else (key,)
    recursive_replace(existing_data, update_path, keys_tuple, value)
```

shrsr (changes requested):

samiib (approved):
LGTM

gmicol (approved):
LGTM

akinross (approved):
LGTM

shrsr (approved):
LGTM

lhercot (approved):
LGTM

PR #550, opened 2024-10-10
[minor_change] Add module ndo_tenant_bgp_peer_prefix_policy for tenant bgp peer prefix policy configuration (DCNE-133)

akinross:
fixes #459

PR #540, opened 2024-09-30
[minor changes] Adding new module for interface policy group object. (DCNE-51)

anvitha-jain:

PR #574, opened 2024-12-02
[minor_change] Added ndo_mcp_global_policy for fabric policy global MCP configuration (DCNE-70)

### LINKED ISSUE #451: Enhancement: New module for MCP Global Policy in Fabric Fabric Policies Template (DCNE-70)
### Community Note

### Description

Create new module for MCP Global Policy in Tenant Fabric Policies Template ( created via ndo_template module )
part of #392 

* mso_XXXXX

* ND 3.1.1 / NDO 4.3.1

### Collection versions

* cisco.mso x.x.x

### References

* #0000

samiib:
Fixed #451 

This MCP Global policy module is based on a 1:1 relationship with Fabric Policy templates.
A decision was made to keep this configuration in a separate module instead of adding it to `ndo_template`.

### DIFF
--- a/plugins/modules/ndo_mcp_global_policy.py
+++ b/plugins/modules/ndo_mcp_global_policy.py
@@ -56,7 +56,8 @@ options:
     type: str
   per_vlan:
     description:
-    - Enable or disable MCP packets being sent to each End Point Group (EPG).
+    - When enabled MCP will send packets on a per End Point Group (EPG) basis.
+    - If disabled, the packets will only be sent on untagged EPGs which allows detecting loops in the native VLAN only.
     - Defaults to C(disabled) when unset during creation.
     type: str
     choices: [ enabled, disabled ]
@@ -65,7 +66,7 @@ options:
     description:
     - The amount of MCP packets that will be received before port disable loop protection action takes place.
     - Defaults to 3 when unset during creation.
-    - The value must be between 0 and 255.
+    - The value must be between 1 and 255.
     type: int
     aliases: [ loop_factor, loop_detection_mult_factor ]
   port_disable:
@@ -121,10 +122,9 @@ EXAMPLES = r"""
     username: admin
     password: SomeSecretPassword
     template: fabric_template
-    name: mcp_global_policy_1
-    key: cisco
+    name: example_mcp_global_policy
     state: present
-  register: mcp_global_policy_1
+  register: mcp_global_policy_new
 
 - name: Create the MCP Global Policy object with all attributes
   cisco.mso.ndo_mcp_global_policy:
@@ -132,7 +132,7 @@ EXAMPLES = r"""
     username: admin
     password: SomeSecretPassword
     template: fabric_template
-    name: mcp_global_policy_1
+    name: example_mcp_global_policy
     description: A Global MCP Policy
     key: cisco
     admin_state: enabled
@@ -143,7 +143,7 @@ EXAMPLES = r"""
     transmission_frequency_sec: 2
     transmission_frequency_msec: 10
     state: present
-  register: mcp_global_policy_1
+  register: mcp_global_policy_all
 
 - name: Update the MCP Global Policy object with UUID
   cisco.mso.ndo_mcp_global_policy:
@@ -152,7 +152,7 @@ EXAMPLES = r"""
     password: SomeSecretPassword
     template: fabric_template
     name: mcp_global_policy_update
-    uuid: "{{ mcp_global_policy_1.current.uuid }}"
+    uuid: "{{ mcp_global_policy_all.current.uuid }}"
     state: present
 
 - name: Query the MCP Global Policy object with name
@@ -161,7 +161,7 @@ EXAMPLES = r"""
     username: admin
     password: SomeSecretPassword
     template: fabric_template
-    name: mcp_global_policy_1
+    name: example_mcp_global_policy
     state: query
   register: query_name
 
@@ -171,7 +171,7 @@ EXAMPLES = r"""
     username: admin
     password: SomeSecretPassword
     template: fabric_template
-    uuid: "{{ mcp_global_policy_1.current.uuid }}"
+    uuid: "{{ mcp_global_policy_all.current.uuid }}"
     state: query
   register: query_uuid
 
@@ -190,7 +190,7 @@ EXAMPLES = r"""
     username: admin
     password: SomeSecretPassword
     template: fabric_template
-    name: mcp_global_policy_1
+    name: example_mcp_global_policy
     state: absent
 
 - name: Delete the MCP Global Policy object with UUID
@@ -199,7 +199,7 @@ EXAMPLES = r"""
     username: admin
     password: SomeSecretPassword
     template: fabric_template
-    uuid: "{{ mcp_global_policy_1.current.uuid }}"
+    uuid: "{{ mcp_global_policy_all.current.uuid }}"
     state: absent
 """
 
@@ -263,9 +263,9 @@ def main():
         "description": module.params.get("description"),
         "adminState": module.params.get("admin_state"),
         "key": module.params.get("key"),
-        "enablePduPerVlan": ENABLED_OR_DISABLED_TO_BOOL_STRING_MAP[per_vlan] if per_vlan else None,
+        "enablePduPerVlan": ENABLED_OR_DISABLED_TO_BOOL_STRING_MAP.get(per_vlan),
         "loopDetectMultFactor": module.params.get("loop_detection_factor"),
-        "protectPortDisable": ENABLED_OR_DISABLED_TO_BOOL_STRING_MAP[port_disable] if port_disable else None,
+        "protectPortDisable": ENABLED_OR_DISABLED_TO_BOOL_STRING_MAP.get(port_disable),
         "initialDelayTime": module.params.get("initial_delay_time"),
         "txFreq": module.params.get("transmission_frequency_sec"),
         "txFreqMsec": module.params.get("transmission_frequency_msec"),
@@ -308,8 +308,8 @@ def main():
 
             mso.sanitize(proposed_payload, collate=True)
         else:
-            if not name:
-                mso.fail_json(msg="{0} name cannot be empty".format(object_description))
+            if uuid:
+                mso.fail_json(msg="{0} cannot be created with a UUID".format(object_description))
             payload = dict()
             for mso_name, mso_value in mso_values.items():
                 if mso_value:
--- a/plugins/modules/ndo_mcp_global_policy.py
+++ b/plugins/modules/ndo_mcp_global_policy.py
@@ -1,0 +1,341 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# Copyright: (c) 2024, Samita Bhattacharjee (@samiib) <samitab@cisco.com>
+
+# GNU General Public License v3.0+ (see LICENSE or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+
+__metaclass__ = type
+
+ANSIBLE_METADATA = {"metadata_version": "1.1", "status": ["preview"], "supported_by": "community"}
+
+DOCUMENTATION = r"""
+---
+module: ndo_mcp_global_policy
+short_description: Manage the MCP Global Policy in a Fabric Policy Template on Cisco Nexus Dashboard Orchestrator (NDO).
+description:
+- Manage the MisCabling Protocol (MCP) Global Policy in a Fabric Policy Template on Cisco Nexus Dashboard Orchestrator (NDO).
+- There can only be a single MCP Global Policy in a Fabric Policy Template.
+- This module is only supported on ND v3.1 (NDO v4.3) and later.
+author:
+- Samita Bhattacharjee (@samiib)
+options:
+  template:
+    description:
+    - The name of the template.
+    - The template must be a Fabric Policy template.
+    type: str
+    aliases: [ fabric_template ]
+    required: true
+  name:
+    description:
+    - The name of the MCP Global Policy.
+    type: str
+    aliases: [ mcp_global_policy ]
+  uuid:
+    description:
+    - The UUID of the MCP Global Policy.
+    type: str
+    aliases: [ mcp_global_policy_uuid ]
+  description:
+    description:
+    - The description of the MCP Global Policy.
+    - Providing an empty string will remove the O(description="") from the MCP Global Policy.
+    type: str
+  admin_state:
+    description:
+    - The administrative state of the MCP Global Policy.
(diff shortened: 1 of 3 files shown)


plugins/modules/ndo_mcp_global_policy.py
```diff
@@ -0,0 +1,341 @@
...
+  loop_detection_factor:
+    description:
+    - The amount of MCP packets that will be received before port disable loop protection action takes place.
+    - Defaults to 3 when unset during creation.
+    - The value must be between 0 and 255.
+    type: int
+    aliases: [ loop_factor, loop_detection_mult_factor ]
+  port_disable:
+    description:
+    - Enable or disable port disabling when MCP packets are recived.
```
shrsr, line 73: -> Enable or disable port disabling when MCP packets are received.

plugins/modules/ndo_mcp_global_policy.py
```diff
@@ -0,0 +1,341 @@
...
+    state: present
+  register: mcp_global_policy_1
+
+- name: Update the MCP Global Policy object with UUID
+  cisco.mso.ndo_mcp_global_policy:
+    host: mso_host
+    username: admin
+    password: SomeSecretPassword
+    template: fabric_template
+    name: mcp_global_policy_1
```
shrsr, line 154: -> name: mcp_global_policy_update
to show that we are updating the name

plugins/modules/ndo_mcp_global_policy.py
```diff
@@ -0,0 +1,341 @@
...
+
+    if state in ["query", "absent"] and existing_mcp_global_policy == {}:
+        mso.exit_json()
+
+    elif state == "query" and not (name or uuid):
+        mso.existing = [existing_mcp_global_policy]
+
+    elif existing_mcp_global_policy and (name or uuid):
+        mso.existing = mso.previous = copy.deepcopy(existing_mcp_global_policy)
+
```
shrsr, line 297: Would it be possible to add check_existing_identifier(mso, object_description, uuid) here, so that we don't have to call it twice? We could even forego the separate function for this if that's possible

shrsr (changes requested):

plugins/modules/ndo_mcp_global_policy.py
samiib: Fixed

plugins/modules/ndo_mcp_global_policy.py
samiib: Done

plugins/modules/ndo_mcp_global_policy.py
samiib: Yes, this is possible. Moved the check logic and removed the function as its only needed in one place now.

plugins/modules/ndo_mcp_global_policy.py
```diff
@@ -0,0 +1,336 @@
...
+    - Defaults to C(enabled) when unset during creation.
+    type: str
+    choices: [ enabled, disabled ]
+  key:
+    description:
+    - The key of the MCP Global Policy.
+    type: str
+  per_vlan:
+    description:
+    - Enable or disable MCP packets being sent to each End Point Group (EPG).
```
akinross, line 59: Noot sure the description fully matches the documentation explanation:

This enables MCP to send packets on a per-EPG basis. If this option is disabled, the packets will only be sent on untagged EPGs which allows detecting loops in the native VLAN only.

plugins/modules/ndo_mcp_global_policy.py
```diff
@@ -0,0 +1,336 @@
...
+    - Defaults to C(disabled) when unset during creation.
+    type: str
+    choices: [ enabled, disabled ]
+    aliases: [ per_epg, mcp_pdu_per_vlan ]
+  loop_detection_factor:
+    description:
+    - The amount of MCP packets that will be received before port disable loop protection action takes place.
+    - Defaults to 3 when unset during creation.
+    - The value must be between 0 and 255.
```
akinross, line 68: 0 gives a UI error

```suggestion
    - The value must be between 1 and 255.
```

plugins/modules/ndo_mcp_global_policy.py
```diff
@@ -0,0 +1,336 @@
...
+    host: mso_host
+    username: admin
+    password: SomeSecretPassword
+    template: fabric_template
+    name: mcp_global_policy_1
+    key: cisco
+    state: present
+  register: mcp_global_policy_1
+
+- name: Create the MCP Global Policy object with all attributes
```
akinross, line 129: Are we adding examples separately with all attributes? Should we do this for all other modules and track as issue?

plugins/modules/ndo_mcp_global_policy.py
```diff
@@ -0,0 +1,336 @@
...
+- name: Create the MCP Global Policy object
+  cisco.mso.ndo_mcp_global_policy:
+    host: mso_host
+    username: admin
+    password: SomeSecretPassword
+    template: fabric_template
+    name: mcp_global_policy_1
+    key: cisco
+    state: present
+  register: mcp_global_policy_1
```
akinross, line 127: do we want the register to be the same value as the example with all attributes?

plugins/modules/ndo_mcp_global_policy.py
```diff
@@ -0,0 +1,336 @@
...
+    state = module.params.get("state")
+    per_vlan = module.params.get("per_vlan")
+    port_disable = module.params.get("port_disable")
+
+    mso_values = {
+        "name": name,
+        "description": module.params.get("description"),
+        "adminState": module.params.get("admin_state"),
+        "key": module.params.get("key"),
+        "enablePduPerVlan": ENABLED_OR_DISABLED_TO_BOOL_STRING_MAP[per_vlan] if per_vlan else None,
```
akinross, line 266: is it possible to leverage the get() function on the dictionary?
```suggestion
        "enablePduPerVlan": ENABLED_OR_DISABLED_TO_BOOL_STRING_MAP.get(per_vlan),
```

plugins/modules/ndo_mcp_global_policy.py
```diff
@@ -0,0 +1,336 @@
...
+        "loopDetectMultFactor": module.params.get("loop_detection_factor"),
+        "protectPortDisable": ENABLED_OR_DISABLED_TO_BOOL_STRING_MAP[port_disable] if port_disable else None,
```
akinross, line 268: is it possible to leverage the get() function on the dictionary?

tests/integration/targets/ndo_mcp_global_policy/tasks/main.yml
```diff
@@ -0,0 +1,300 @@
...
+
+    - name: Query the MCP Global Policy with UUID
+      cisco.mso.ndo_mcp_global_policy:
+        <<: *mso_info
+        template: ansible_fabric_policy_template
+        uuid: '{{update_mcp_global_policy_again.current.uuid}}'
+        state: query
+      register: query_mcp_policy_with_uuid
+
+    - name: Query the MCP Global Policy without name or UUID
```
akinross, line 180: I think in other we have been doing a query with both name and UUID provided, could you add that?

tests/integration/targets/ndo_mcp_global_policy/tasks/main.yml
```diff
@@ -0,0 +1,300 @@
...
+          - nm_delete_mcp_global_policy_uuid.previous.description == "An MCP Global policy to delete with UUID"
+          - nm_delete_mcp_global_policy_uuid.current == {}
+
+    # Errors and no policy found
+    - name: Query the MCP Global Policy in the template when its deleted
+      cisco.mso.ndo_mcp_global_policy:
+        <<: *query_mcp_global_policy
+      register: query_none
+
+    - name: Create an MCP Global Policy without a name
```
akinross, line 278: Is this test input correct, because I would argue that a create with UUID is not possible because the policy is not created yet. So uuid is not needed to be provided?

tests/integration/targets/ndo_mcp_global_policy/tasks/main.yml
```diff
@@ -0,0 +1,300 @@
...
+          - nm_delete_mcp_global_policy_again.previous == nm_delete_mcp_global_policy_again.current == {}
+
+    - name: Create the MCP Global Policy
+      cisco.mso.ndo_mcp_global_policy:
+        <<: *add_mcp_global_policy
+        name: ansible_mcp_global_policy_again
+        description: An MCP Global policy to delete with UUID
+      register: new_mcp_global_policy
+
+    - name: Delete the MCP Global Policy with incorrect UUID
```
akinross, line 246: do we want to move this test to the errors section?

tests/integration/targets/ndo_mcp_global_policy/tasks/main.yml
```diff
@@ -0,0 +1,300 @@
...
+      register: query_none
+
+    - name: Create an MCP Global Policy without a name
+      cisco.mso.ndo_mcp_global_policy:
+        <<: *mso_info
+        template: ansible_fabric_policy_template
+        uuid: non-existing-uuid
+        state: present
+      ignore_errors: true
+      register: update_non_existing_uuid
```
akinross, line 285:
```suggestion
      register: create_without_name
```

plugins/modules/ndo_mcp_global_policy.py
samiib: The enabled description sounds like it matches to me. However, I will add the information when the option is disabled.
Would it be okay to share the link to the documentation?

plugins/modules/ndo_mcp_global_policy.py
samiib: I had seen an example with all or most attributes in a [previous PR](https://github.com/CiscoDevNet/ansible-mso/pull/540/files#r1812032660) and thought it was a good idea.
We can add a new issue to do it for all modules if you would like to?

akinross: I like the idea but think we should track if we deviate from what is there in other modules. I missed this in that PR but think that is the only PR where it is done that way, which introduces a inconsistency from the rest.

plugins/modules/ndo_mcp_global_policy.py
akinross: Ok perhaps it is my perception that "to each EPG" vs "a per-EPG basis" is slightly different. So never mind if that is the same, would add the disabled text.

What do you mean with link to the documentation? In the description a link to https://www.cisco.com/c/en/us/td/docs/dcn/ndo/4x/configuration/cisco-nexus-dashboard-orchestrator-configuration-guide-aci-421/ndo-configuration-aci-fabric-management-42x.html? Or as general in the see also part?

tests/integration/targets/ndo_mcp_global_policy/tasks/main.yml
samiib: Would it be okay to link to an example test that queries with both name and UUID?
The closest I can see is [this recent one](https://github.com/CiscoDevNet/ansible-mso/pull/550/files#diff-643a3396f501aad132095eb0e57a7b43569d3241d48b1cea7cb77b740013d3b7R228) which uses `name: '{{ fakevar | default(omit)}}'`, since `fakevar` doesn't exist it will omit name.

plugins/modules/ndo_mcp_global_policy.py
samiib: I was just curious to see the documentation you referred to, not to add it to the module.

tests/integration/targets/ndo_mcp_global_policy/tasks/main.yml
samiib: Done

plugins/modules/ndo_mcp_global_policy.py
samiib: Done

plugins/modules/ndo_mcp_global_policy.py
samiib: Done

plugins/modules/ndo_mcp_global_policy.py
samiib: Changed

tests/integration/targets/ndo_mcp_global_policy/tasks/main.yml
samiib: Done

plugins/modules/ndo_mcp_global_policy.py
samiib: Created DCNE-265

tests/integration/targets/ndo_mcp_global_policy/tasks/main.yml
akinross: Ok think this is the one that I meant, so not relevant in that case.

akinross (approved):
LGTM

shrsr (dismissed):
LGTM

gmicol (dismissed):
LGTM

sajagana (approved):
LGTM!

sajagana (approved):
LGTM!

plugins/modules/ndo_mcp_global_policy.py
```diff
@@ -0,0 +1,323 @@
...
+
+    mso = MSOModule(module)
+
+    name = module.params.get("name")
+    uuid = module.params.get("uuid")
+    state = module.params.get("state")
+    per_vlan = module.params.get("per_vlan")
+    port_disable = module.params.get("port_disable")
+
+    mso_values = {
```
akinross, line 262: shall we move this to present, since it is not used anywhere else?

akinross: should we use dict() or {}? we use dict() most of the time for dict instantiation, but not sure if it matters

akinross (changes requested):

plugins/modules/ndo_mcp_global_policy.py
samiib: I'll keep it here for now. It keeps all the `module.params.get` in the same place.

There appears to be a difference in performance. Eg https://doughellmann.com/posts/the-performance-impact-of-using-dict-instead-of-in-cpython-2-7-2/
The conclusion looks like `{}` is more efficient. Although this post is for 2.7, I can still observe a similar result in 3.

shrsr: I think we have always used dict() to keep the code consistent with how we integrate it with Ansible.
Example:
```
argument_spec.update(
        template=dict(type="str", required=True, aliases=["fabric_template"]),
```
If we decide to go with {} for the payload then we would likely need to change it everywhere.

akinross (approved):
LGTM

anvitha-jain (approved):

lhercot (approved):
LGTM

PR #578, opened 2024-12-10
Addition of new module ndo_ptp_policy and its child ndo_ptp_policy_profiles (DCNE-68)

shrsr:


plugins/modules/ndo_ptp_policy.py
```diff
@@ -0,0 +1,297 @@
...
+    global_domain: 0
+    fabric_delay_interval: -2
+    global_priority1: 255
+    global_priority2: 255
+    fabric_announce_timeout: 3
+    fabric_announce_interval: 1
+    fabric_profile_template: aes67
+    state: present
+
+- name: Query a PTP policy with ptp_policy name
```
akinross, line 127: the query and create examples with uuid are missing

plugins/modules/ndo_ptp_policy.py
```diff
@@ -0,0 +1,297 @@
...
+__metaclass__ = type
+
+ANSIBLE_METADATA = {"metadata_version": "1.1", "status": ["preview"], "supported_by": "community"}
+
+DOCUMENTATION = r"""
+---
+module: ndo_ptp_policy
+short_description: Manage PTP Policies on Cisco Nexus Dashboard Orchestrator (NDO).
+description:
+- Manage PTP Policies on Cisco Nexus Dashboard Orchestrator (NDO).
```
akinross, line 19: think we did 1 time abbreviation, should we add here?
```suggestion
- Manage Precision Time Protocol (PTP) Policies on Cisco Nexus Dashboard Orchestrator (NDO).
```

plugins/modules/ndo_ptp_policy.py
```diff
@@ -0,0 +1,297 @@
...
+    description:
+    - Fabric announce interval for the PTP policy.
+    - The value should be between 0 to 4.
+    - This is required when O(state=present).
+    type: int
+  fabric_profile_template:
+    description:
+    - Fabric profile template for the PTP policy.
+    type: str
+    choices: [ aes67, default, smpte ]
```
akinross, line 96: should we match these to the complete UI inputs?

aes67-2015 and smpte-2059-2

akinross (changes requested):

plugins/modules/ndo_ptp_policy_profiles.py
```diff
@@ -0,0 +1,275 @@
...
+  ptp_policy_profile_uuid:
+    description:
+    - The UUID of the PTP Policy profile.
+    type: str
+    aliases: [ uuid ]
+  delay_interval:
+    description:
+    - Delay interval of the PTP Policy profile.
+    - The value should be between -4 to 4.
+    - This is required when O(state=present).
```
gmicol, line 44: Are attributes like `delay_interval` which are described as `required when O(state=present)` needed for creation only or everytime the user manipulates a PTP Policy Profile (creation + update)?

If creation only, does the API error out on its own? if not, we should create flags to check if the attributes exist but only for creation steps.

And the same thing applies for ndo_ptp_policy?

gmicol (changes requested):

plugins/modules/ndo_ptp_policy_profiles.py
shrsr: When it comes to the module ndo_ptp_policy and we take the two below configs as example where fabric_delay_interval is missing from the second config.
```
   - name: 1st config
      cisco.mso.ndo_ptp_policy: 
        <<: *mso_info
        template: ansible_fabric_policy_template
        ptp_policy: ansible_test_ptp_policy
        fabric_sync_interval: -3
        global_domain: 0
        fabric_delay_interval: -2
        global_priority1: 255
        global_priority2: 255
        fabric_announce_timeout: 3
        fabric_announce_interval: 1
        fabric_profile_template: aes67-2015
        state: present
        
   - name: 2nd config
      cisco.mso.ndo_ptp_policy: 
        <<: *mso_info
        template: ansible_fabric_policy_template
        ptp_policy: ansible_test_ptp_policy
        fabric_sync_interval: -3
        global_domain: 0
        global_priority1: 255
        global_priority2: 255
        fabric_announce_timeout: 3
        fabric_announce_interval: 1
        fabric_profile_template: aes67-2015
        state: present
 ```
Technically the 2nd config should show as "no change" but it does for this module because the payload has an extra key "global" for a few sub-attributes and their values as seen below:
```
mso_values = {
            "name": ptp_policy,
            "description": description,
            "global": {
                "adminState": admin_state,
                "fabSyncIntvl": fabric_sync_interval,
                "globalDomain": global_domain,
                "fabDelayIntvl": fabric_delay_interval,
                "prio1": global_priority1,
                "prio2": global_priority2,
                "fabAnnounceTimeout": fabric_announce_timeout,
                "fabAnnounceIntvl": fabric_announce_interval,
                "fabProfileTemplate": fabric_profile_template,
            },
        }
 ```
When we don't include fabric_delay_interval above, the new append_update_ops_data() actually detects a change in the entire value of global and not fabDelayIntvl specifically. As a result a null value is passed to fabDelayIntvl which NDO consumes as "0" and this is actually a valid value for fabDelayIntvl.

For now I added this to prevent a change which enforces the user to actually specify it every time.
```
[
                "state",
                "present",
                [
                    "global_priority1",
                    "global_priority2",
                    "global_domain",
                    "fabric_announce_timeout",
                    "fabric_announce_interval",
                    "fabric_sync_interval",
                    "fabric_delay_interval",
                ],
            ],
```
What we can do is not use append_update_ops_data() at all for this particular module and have individual checks like before.

plugins/modules/ndo_ptp_policy_profiles.py
```diff
@@ -0,0 +1,275 @@
...
+    type: int
+  state:
+    description:
+    - Use C(absent) for removing.
+    - Use C(query) for listing an object or multiple objects.
+    - Use C(present) for creating or updating.
+    type: str
+    choices: [ absent, query, present ]
+    default: query
```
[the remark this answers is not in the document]
shrsr: I think it's query for all the new ndo modules. This will need to change though.

plugins/modules/ndo_ptp_policy.py
```diff
@@ -0,0 +1,329 @@
...
+    type: str
+    choices: [ aes67-2015, default, smpte-2059-2 ]
+  state:
+    description:
+    - Use C(absent) for removing.
+    - Use C(query) for listing an object or multiple objects.
+    - Use C(present) for creating or updating.
+    type: str
+    choices: [ absent, query, present ]
+    default: query
```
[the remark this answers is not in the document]
shrsr: We would need to change this throughout the new ndo modules..

plugins/modules/ndo_ptp_policy.py
```diff
@@ -0,0 +1,334 @@
...
+            [existing_ptp_policies],
+            [KVPair("uuid", ptp_policy_uuid) if ptp_policy_uuid else KVPair("name", ptp_policy)],
+        )
+        if match:
+            mso.existing = mso.previous = copy.deepcopy(match.details)
+    else:
+        mso.existing = mso.previous = existing_ptp_policies
+
+    if state == "present":
+        mso_values_update = {
```
sajagana, line 275: Use `dict()` instead of `{}` and move it under `if match:`.

plugins/modules/ndo_ptp_policy.py
```diff
@@ -0,0 +1,334 @@
...
+            ("global", "fabAnnounceTimeout"): fabric_announce_timeout,
+            ("global", "fabAnnounceIntvl"): fabric_announce_interval,
+            ("global", "fabProfileTemplate"): fabric_profile_template,
+        }
+
+        if match:
+            append_update_ops_data(ops, match.details, path, mso_values_update)
+            mso.sanitize(match.details, collate=True)
+        else:
+            mso_values = {
```
sajagana, line 293: Use dict() instead of {}.

plugins/module_utils/constants.py
```diff
@@ -211,3 +211,5 @@
 
 ORIGINATE_DEFAULT_ROUTE = {"only": "only", "in_addition": "inAddition", "": ""}
 L3OUT_ROUTING_PROTOCOLS = {"bgp": ["bgp"], "ospf": ["ospf"], "bgpOspf": ["bgp", "ospf"], None: [None], "": None, "bgpospf": "bgpOspf", "ospfbgp": "bgpOspf"}
+
+PROFILE_TEMPLATE = {"aes67-2015": "aes67", "default": "default", "smpte-2059-2": "smpte", "telecom-8275-1": "telecomFullPath"}
```
sajagana, line 215: Can we use `_` instead of `-`?

plugins/modules/ndo_ptp_policy_profiles.py
```diff
@@ -0,0 +1,278 @@
...
+
+def main():
+    argument_spec = mso_argument_spec()
+    argument_spec.update(
+        dict(
+            template=dict(type="str", required=True),
+            ptp_policy_profile_name=dict(type="str", aliases=["name"]),
+            ptp_policy_profile_uuid=dict(type="str", aliases=["uuid"]),
+            delay_interval=dict(type="int"),
+            profile_template=dict(type="str", choices=["aes67-2015", "default", "telecom-8275-1", "smpte-2059-2"]),
```
sajagana, line 182:
```suggestion
            profile_template=dict(type="str", choices=list(PROFILE_TEMPLATE)),
```

sajagana (changes requested):

plugins/modules/ndo_ptp_policy.py
shrsr: Moved under match. I didn't change it to dict() in order to be consistent with mso_values. Check comment!

plugins/modules/ndo_ptp_policy.py
shrsr: I went with {} because one of the keys is "global" which is a python keyword. Python complained when I used "global" as a key inside dict()

samiib (changes requested):

tests/integration/targets/ndo_ptp_policy/tasks/main.yml
```diff
@@ -0,0 +1,233 @@
...
+        template: ansible_fabric_policy_template
+        state: query
+      register: query_all
+
+    - name: Verify query_one and query_all
+      assert:
+        that:
+          - query_all is not changed
+          - query_one is not changed
+          - query_all.current == query_one.current
```
samiib, line 184: Should we add another PTP policy to test querying more than 1 policy?

tests/integration/targets/ndo_ptp_policy/tasks/main.yml
```diff
@@ -0,0 +1,233 @@
...
+          - nm_update_ptp_policy.current.description == "changed_description"
+          - nm_update_ptp_policy.current.global.adminState == "disabled"
+          - nm_update_ptp_policy.current.global.fabSyncIntvl == -2
+          - nm_update_ptp_policy.current.global.globalDomain == 1
+          - nm_update_ptp_policy.current.global.prio1 == 100
+          - nm_update_ptp_policy.current.global.prio2 == 200
+          - nm_update_ptp_policy_again is not changed
+
+    # QUERY
+    - name: Update PTP policy name with UUID
```
samiib, line 157: Should the update with UUID task be in the update test section above and also assert?

tests/integration/targets/ndo_ptp_policy/tasks/main.yml
```diff
@@ -0,0 +1,233 @@
...
+      cisco.mso.ndo_ptp_policy:
+        <<: *update_ptp_policy
+        ptp_policy: ansible_test_ptp_policy_updated
+        ptp_policy_uuid: "{{ nm_add_ptp_policy.current.uuid }}"
+        state: present
+
+    - name: Query a PTP policy with ptp_policy name
```
samiib, line 164: Would it be possible to add a test to query by UUID?

tests/integration/targets/ndo_ptp_policy/tasks/main.yml
```diff
@@ -0,0 +1,233 @@
...
+          - query_one.current.description == "changed_description"
+          - query_one.current.global.adminState == "disabled"
+          - query_one.current.global.fabSyncIntvl == -2
+          - query_one.current.global.globalDomain == 1
+          - query_one.current.global.prio1 == 100
+          - query_one.current.global.prio2 == 200
+          - query_one.current.global.fabAnnounceTimeout == 3
+          - query_one.current.global.fabAnnounceIntvl == 1
+          - query_one.current.global.fabProfileTemplate == "aes67"
+
```
samiib, line 195: Are there any ERROR test cases that can be added here? Potentially some testing the value ranges?

tests/integration/targets/ndo_ptp_policy_profiles/tasks/main.yml
```diff
@@ -0,0 +1,252 @@
...
+      - query_all is not changed
+      - query_one is not changed
+      - nm_update_name_ptp_profile.current.name == "profile_update"
+      - query_one.current.name == "profile_update"
+      - query_one.current.delayIntvl == 5
+      - query_one.current.syncIntvl == -1
+      - query_one.current.announceTimeout == 6
+      - query_one.current.announceIntvl == 3
+      - query_one.current.profileTemplate == "default"
+      - query_all.current | length == 2
```
samiib, line 206: Could we also check the two profiles returned are correct? Checking the name should be enough.

tests/integration/targets/ndo_ptp_policy_profiles/tasks/main.yml
```diff
@@ -0,0 +1,252 @@
...
+
```
samiib, line 207: Are there any ERROR test cases that can be added?

samiib (changes requested):

tests/integration/targets/ndo_ptp_policy/tasks/main.yml
shrsr: Only one PTP policy can be added

tests/integration/targets/ndo_ptp_policy_profiles/tasks/main.yml
shrsr: There are none

tests/integration/targets/ndo_ptp_policy/tasks/main.yml
shrsr: The API returns an error message if the value passed is not in range and our code doesn't take of this, so no tests are necessary here

gmicol (dismissed):
LGTM

anvitha-jain (approved):
LGTM

sajagana (dismissed):
LGTM!

samiib (approved):
LGTM

akinross (dismissed):
LGTM

lhercot (approved):
LGTM

PR #577, opened 2024-12-09
Adding modules for IGMP interface and tenant route map policy object.

anvitha-jain:

PR #600, opened 2025-01-22
[minor_change] Added new module for IPSLA Track Lists (DCNE-134)

### LINKED ISSUE #460: Enhancement: New module for IPSLA Track List Policy in Tenant Tenant Policies Template (DCNE-134)
### Community Note

### Description

Create new module for IPSLA Track List Policy in Tenant Tenant Policies Template ( created via ndo_template module )

* mso_XXXXX

* ND 3.1.1 / NDO 4.3.1

### Collection versions

* cisco.mso x.x.x

### References

* #0000

samiib:
Fixes #460

### DIFF
--- a/plugins/module_utils/schema.py
+++ b/plugins/module_utils/schema.py
@@ -15,24 +15,23 @@ Item = namedtuple("Item", "index details")
 
 
 class MSOSchema:
-    def __init__(self, mso_module, schema_name, template_name=None, site_name=None):
+    def __init__(self, mso_module, schema_name, template_name=None, site_name=None, schema_id=None, template_id=None):
         self.mso = mso_module
         self.schema_name = schema_name
-        self.id, self.path, self.schema = mso_module.query_schema(schema_name)
+        if schema_id:
+            self.id, self.path, self.schema = mso_module.query_schema_by_id(schema_id)
+            self.schema_name = self.schema.get("displayName")
+        else:
+            self.id, self.path, self.schema = mso_module.query_schema(schema_name)
         self.schema_objects = {}
-        if template_name:
-            self.set_template(template_name)
-        if site_name and template_name:
-            self.set_site(template_name, site_name)
-
-    @classmethod
-    def with_template_id(cls, mso_module, schema_name, template_name=None, template_id=None):
-        schema = cls(mso_module, schema_name)
-        if template_id:
-            schema.set_template_from_id(template_id)
-        elif template_name:
-            schema.set_template(template_name)
-        return schema
+        self.template_id = template_id
+        self.template_name = template_name
+        if self.template_id:
+            self.set_template_from_id(self.template_id)
+        elif self.template_name:
+            self.set_template(self.template_name)
+        if site_name and self.template_name:
+            self.set_site(self.template_name, site_name)
 
     @staticmethod
     def get_object_from_list(search_list, kv_list):
@@ -75,6 +74,7 @@ class MSOSchema:
         if not match and fail_module:
             msg = "Provided template '{0}' not matching existing template(s): {1}".format(template_name, ", ".join(existing))
             self.mso.fail_json(msg=msg)
+        self.template_id = match.details.get("templateID")
         self.schema_objects["template"] = match
 
     def set_template_from_id(self, template_id, fail_module=True):
@@ -89,6 +89,7 @@ class MSOSchema:
         if not match and fail_module:
             msg = "Provided template ID '{0}' not matching existing template(s): {1}".format(template_id, ", ".join(existing))
             self.mso.fail_json(msg=msg)
+        self.template_name = match.details.get("name")
         self.schema_objects["template"] = match
 
     def set_template_vrf(self, vrf, fail_module=True):
--- a/plugins/module_utils/schema.py
+++ b/plugins/module_utils/schema.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 
 # Copyright: (c) 2022, Akini Ross (@akinross) <akinross@cisco.com>
+# Copyright: (c) 2025, Samita Bhattacharjee (@samiib) <samitab@cisco.com>
 # GNU General Public License v3.0+ (see LICENSE or https://www.gnu.org/licenses/gpl-3.0.txt)
 
 from __future__ import absolute_import, division, print_function
@@ -23,6 +24,15 @@ class MSOSchema:
             self.set_template(template_name)
         if site_name and template_name:
             self.set_site(template_name, site_name)
+
+    @classmethod
+    def with_template_id(cls, mso_module, schema_name, template_name=None, template_id=None):
+        schema = cls(mso_module, schema_name)
+        if template_id:
+            schema.set_template_from_id(template_id)
+        elif template_name:
+            schema.set_template(template_name)
+        return schema
 
     @staticmethod
     def get_object_from_list(search_list, kv_list):
@@ -64,6 +74,20 @@ class MSOSchema:
         match, existing = self.get_object_from_list(self.schema.get("templates"), kv_list)
         if not match and fail_module:
             msg = "Provided template '{0}' not matching existing template(s): {1}".format(template_name, ", ".join(existing))
+            self.mso.fail_json(msg=msg)
+        self.schema_objects["template"] = match
+
+    def set_template_from_id(self, template_id, fail_module=True):
+        """
+        Get template item that matches the id of a template.
+        :param template_id: ID of the template to match. -> Str
+        :param fail_module: When match is not found fail the ansible module. -> Bool
+        :return: Template item. -> Item(Int, Dict) | None
+        """
+        kv_list = [KVPair("templateID", template_id)]
+        match, existing = self.get_object_from_list(self.schema.get("templates"), kv_list)
+        if not match and fail_module:
+            msg = "Provided template ID '{0}' not matching existing template(s): {1}".format(template_id, ", ".join(existing))
             self.mso.fail_json(msg=msg)
         self.schema_objects["template"] = match
 
--- a/plugins/module_utils/template.py
+++ b/plugins/module_utils/template.py
@@ -461,6 +461,11 @@ class MSOTemplate:
                         config_data[reference_details.get("template")] = template_object.get("templateName")
                     if reference_details.get("templateId"):
                         config_data[reference_details.get("templateId")] = template_object.get("templateId")
+                    if reference_details.get("schemaId"):
+                        config_data[reference_details.get("schemaId")] = template_object.get("schemaId")
+                    if reference_details.get("schema"):
+                        config_data[reference_details.get("schema")] = template_object.get("schemaName")
+            return config_data
         return config_data
 
     def check_template_when_name_is_provided(self, parameter):
--- a/plugins/module_utils/template.py
+++ b/plugins/module_utils/template.py
@@ -27,6 +27,7 @@ class MSOTemplate:
         self.template_id = template_id
         self.template_type = template_type
         self.template_summary = {}
+        self.template_objects_cache = {}
 
         if template_id:
             # Checking if the template with id exists to avoid error: MSO Error 400: Template ID 665da24b95400f375928f195 invalid
@@ -328,6 +329,9 @@ class MSOTemplate:
         match = self.get_object_by_key_value_pairs(policy_type, existing_policies, [KVPair("name", policy_name)], fail_module=True)
         return match.details.get("uuid")
 
+    def clear_template_objects_cache(self):
+        self.template_objects_cache = {}
+
     def get_template_object_name_by_uuid(self, object_type, uuid, fail_module=True):
         """
         Retrieve the name of a specific object type in the MSO template using its UUID.
@@ -342,26 +346,34 @@ class MSOTemplate:
         if response_object:
             return response_object.get("name")
 
-    def get_template_object_by_uuid(self, object_type, uuid, fail_module=True):
+    def get_template_object_by_uuid(self, object_type, uuid, fail_module=True, use_cache=False):
         """
         Retrieve a specific object type in the MSO template using its UUID.
-        :param object_type: The type of the object to retrieve the name for -> Str
-        :param uuid: The UUID of the object to retrieve the name for -> Str
+        :param object_type: The type of the object to retrieve -> Str
+        :param uuid: The UUID of the object to retrieve -> Str
+        :param use_cache: Use the cached result of the templates/objects API for the UUID -> Bool
         :return: Dict | None: The processed result which could be:
             When the UUID is existing, returns object -> Dict
             When the UUID is not existing -> None
         """
-        response_object = self.mso.request("templates/objects?type={0}&uuid={1}".format(object_type, uuid), "GET")
+        response_object = None
+        if use_cache and uuid in self.template_objects_cache.keys():
+            response_object = self.template_objects_cache[uuid]
+        else:
+            response_object = self.mso.request("templates/objects?type={0}&uuid={1}".format(object_type, uuid), "GET")
+            self.template_objects_cache[uuid] = response_object
         if not response_object and fail_module:
             msg = "Provided {0} with UUID of '{1}' not found.".format(object_type, uuid)
             self.mso.fail_json(msg=msg)
         return response_object
 
-    def update_config_with_template_and_references(self, config_data, reference_collections=None):
+    def update_config_with_template_and_references(self, config_data, reference_collections=None, set_template=True, use_cache=False):
         """
         Return the updated config_data with the template values and reference_collections if provided
         :param config_data: The original config_data that requires to be updated -> Dict
         :param reference_collections: A dict containing the object type, references and the corresponding names -> Dict
+        :param set_template: Adds the templateId and templateName to the config_data -> Bool
+        :param use_cache: Use the cached result of the templates/objects API for the ref UUID -> Bool
         :return: Updated config_data with names for references -> Dict
         Example 1:
         reference_collections = {
@@ -431,16 +443,19 @@ class MSOTemplate:
         """
 
         # Set template ID and template name if available
-        if self.template_id:
-            config_data["templateId"] = self.template_id
-        if self.template_name:
-            config_data["templateName"] = self.template_name
+        if set_template:
+            if self.template_id:
+                config_data["templateId"] = self.template_id
+            if self.template_name:
+                config_data["templateName"] = self.template_name
 
         # Update config data with reference names if reference_collections is provided
         if reference_collections:
             for reference_details in reference_collections.values():
                 if config_data.get(reference_details.get("reference")):
-                    template_object = self.get_template_object_by_uuid(reference_details.get("type"), config_data.get(reference_details.get("reference")))
+                    template_object = self.get_template_object_by_uuid(
+                        reference_details.get("type"), config_data.get(reference_details.get("reference")), True, use_cache
+                    )
                     config_data[reference_details.get("name")] = template_object.get("name")
                     if reference_details.get("template"):
                         config_data[reference_details.get("template")] = template_object.get("templateName")
--- a/plugins/module_utils/template.py
+++ b/plugins/module_utils/template.py
@@ -221,6 +221,28 @@ class MSOTemplate:
         match = self.get_object_by_key_value_pairs("Interface Policy Groups", existing_policy_groups, kv_list, fail_module=True)
         return match.details.get("uuid")
 
+    def get_ipsla_monitoring_policy(self, uuid=None, name=None, fail_module=False):
+        """
+        Get the IPSLA Monitoring Policy by UUID or Name.
+        :param uuid: UUID of the IPSLA Monitoring Policy to search for -> Str
+        :param name: Name of the IPSLA Monitoring Policy to search for -> Str
+        :param fail_module: When match is not found fail the ansible module -> Bool
+        :return: Dict | None | List[Dict] | List[]: The processed result which could be:
+                 When the UUID | Name is existing in the search list -> Dict
+                 When the UUID | Name is not existing in the search list -> None
+                 When both UUID and Name are None, and the search list is not empty -> List[Dict]
+                 When both UUID and Name are None, and the search list is empty -> List[]
+        """
+        existing_ipsla_policies = self.template.get("tenantPolicyTemplate", {}).get("template", {}).get("ipslaMonitoringPolicies", [])
+        if name or uuid:
+            return self.get_object_by_key_value_pairs(
+                "IPSLA Monitoring Policy",
+                existing_ipsla_policies,
+                [(KVPair("uuid", uuid) if uuid else KVPair("name", name))],
+                fail_module=fail_module,
+            )
+        return existing_ipsla_policies
+
     def get_l3out_object(self, uuid=None, name=None, fail_module=False):
         """
         Get the L3Out by uuid or name.
--- a/plugins/module_utils/template.py
+++ b/plugins/module_utils/template.py
@@ -238,7 +238,7 @@ class MSOTemplate:
             return self.get_object_by_key_value_pairs(
                 "IPSLA Monitoring Policy",
                 existing_ipsla_policies,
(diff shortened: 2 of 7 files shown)


plugins/modules/ndo_ipsla_track_list.py
```diff
@@ -0,0 +1,536 @@
...
+        if thresholds["down"] > thresholds["up"]:
+            mso.fail_json(
+                msg="Invalid value provided for threshold_down: {0}; it must be less than or equal to threshold_up: {1}".format(
+                    thresholds["down"], thresholds["up"]
+                )
+            )
+
+    ops = []
+    match = None
+    obj_cache = {}
```
samiib, line 377: The `obj_cache` is used so each `member` does not need to send an API call to get schema/template details multiple times.

plugins/modules/ndo_ipsla_track_list.py
```diff
@@ -0,0 +1,536 @@
...
+                type="list",
+                elements="dict",
+                options=dict(
+                    destination_ip=dict(type="str", aliases=["ip"], required=True),
+                    ipsla_monitoring_policy_uuid=dict(type="str"),
+                    ipsla_monitoring_policy=dict(
+                        type="dict",
+                        options=dict(
+                            name=dict(type="str", required=True),
+                            template=dict(type="str", required=True),
```
shrsr, line 302: Is template really required here?
Can you check if an IPSLA that belongs to a different template can be passed here?

plugins/modules/ndo_ipsla_track_list.py
```diff
@@ -0,0 +1,536 @@
...
+
+        if type == "l3out":
+            if not mso_template:
+                mso_template = MSOTemplate(mso, type, template)
+                mso_template.validate_template(type)
+                obj_cache[key] = mso_template
+            return get_l3out_uuid(mso_template, name)
+
+        elif type == "tenant":
+            if not mso_template:
```
shrsr, line 509: If template is not required as in if NDO doesn't allow the user to associate an IPSLA from a different template then we can directly use mso_template which was initialized the initial instantiation of MSOTemplate in line 379

shrsr (changes requested):

anvitha-jain (dismissed):
LGTM

plugins/modules/ndo_ipsla_track_list.py
samiib: I recall being able to add an IPLSA Monitoring Policy from a different template. I will double check.

samiib: Yes, you are right. The UI only lets you select a IPSLA Monitoring Policy from the same template. Although, BD and L3Out scope can be from different templates. 👍 
I'll change it.

samiib: Done

plugins/modules/ndo_ipsla_track_list.py
samiib: Done

plugins/modules/ndo_ipsla_track_list.py
```diff
@@ -0,0 +1,511 @@
...
+            "destIP": member.get("destination_ip"),
+            "scope": get_scope_obj_uuid(scope_type, member.get("scope_uuid"), member.get("scope")),
+            "scopeType": scope_type,
+            "ipslaMonitoringRef": get_ipsla_monitoring_policy_uuid(
+                mso_template,
+                member.get("ipsla_monitoring_policy_uuid"),
+                member.get("ipsla_monitoring_policy"),
+            ),
+        }
+        track_member = {"trackMember": track_member}
```
akinross, line 502: why do this in a two step approach?
```suggestion
 track_member = {
    "trackMember": {
      "destIP": member.get("destination_ip"),
      "scope": get_scope_obj_uuid(scope_type, member.get("scope_uuid"), member.get("scope")),
      "scopeType": scope_type,
      "ipslaMonitoringRef": get_ipsla_monitoring_policy_uuid(
          mso_template,
          member.get("ipsla_monitoring_policy_uuid"),
          member.get("ipsla_monitoring_policy"),
      ),
    }
  }
```

akinross (changes requested):

plugins/modules/ndo_ipsla_track_list.py
samiib: I think its a good idea for a future enhancement. We could potentially explore different libraries that may help. eg https://pypi.org/project/cachetools/
However, it would be quite a bit of work. It may also only be useful on modules with lists of dictionaries that need to call APIs for each item, like this module and `mso_schema_template_anp_epg_contract`. A simple dictionary cache like this is straight forward enough to implement on a case-by-case basis for now.

plugins/modules/ndo_ipsla_track_list.py
```diff
@@ -0,0 +1,511 @@
...
+    if state == "present":
+        if ipsla_track_list_uuid and not mso.existing:
+            mso.fail_json(msg="{0} with the UUID: '{1}' not found".format(object_description, ipsla_track_list_uuid))
+
+        mso_values = {
+            "name": ipsla_track_list,
+            "description": description,
+            "type": type,
+            type + "Up": thresholds["up"],
+            type + "Down": thresholds["down"],
```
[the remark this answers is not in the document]
samiib: It looked a little more readable to me. I'll change it though.

plugins/modules/ndo_ipsla_track_list.py
```diff
@@ -0,0 +1,511 @@
...
+        scope:
+          name: ansible_test_l3out
+          template: ansible_test_template
+        ipsla_monitoring_policy: ansible_test_ipsla_monitoring_policy
+      - destination_ip: 1.1.1.2
+        scope_type: l3out
+        scope_uuid: "{{ l3out.current.uuid }}"
+        ipsla_monitoring_policy_uuid: "{{ ipsla_monitoring_policy.current.uuid }}"
+    state: present
+
```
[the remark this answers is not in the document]
samiib: Added

plugins/modules/ndo_ipsla_track_list.py
```diff
@@ -0,0 +1,511 @@
...
+            "name": ipsla_track_list,
+            "description": description,
+            "type": type,
+            type + "Up": thresholds["up"],
+            type + "Down": thresholds["down"],
+        }
+        if members is not None:
+            mso_values["trackListMembers"] = format_track_list_members(mso, mso_template, members, obj_cache)
+        if match:
+            update_path = "{0}/{1}".format(path, match.index)
```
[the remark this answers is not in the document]
samiib: Moved this inline.

plugins/modules/ndo_ipsla_track_list.py
```diff
@@ -0,0 +1,511 @@
...
+            "ipslaMonitoringRef": get_ipsla_monitoring_policy_uuid(
+                mso_template,
+                member.get("ipsla_monitoring_policy_uuid"),
+                member.get("ipsla_monitoring_policy"),
+            ),
+        }
+        track_member = {"trackMember": track_member}
+        weight = member.get("weight")
+        if weight is not None:
+            track_member["weight"] = weight
```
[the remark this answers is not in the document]
samiib: Sanitise is cleaning it, fixed.

plugins/modules/ndo_ipsla_track_list.py
samiib: Changed

plugins/modules/ndo_ipsla_track_list.py
```diff
@@ -0,0 +1,511 @@
...
+def get_ipsla_monitoring_policy_uuid(tenant_template_obj, uuid, name):
+    existing_ipsla_policies = tenant_template_obj.template.get("tenantPolicyTemplate", {}).get("template", {}).get("ipslaMonitoringPolicies", [])
+    match = tenant_template_obj.get_object_by_key_value_pairs(
+        "IPSLA Monitoring Policy",
+        existing_ipsla_policies,
+        [(KVPair("uuid", uuid) if uuid else KVPair("name", name))],
+        fail_module=True,
+    )
+    if match:
+        return match.details.get("uuid")
```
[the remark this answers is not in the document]
samiib: I have updated it to use existing functions for BD and L3Out objects. I added a new function for IPSLA Monitoring policies.

plugins/modules/ndo_ipsla_track_list.py
```diff
@@ -0,0 +1,490 @@
...
+        description:
+        - The UUID of the IPSLA Monitoring Policy to use for the member.
+        - This parameter can be used instead of O(members.ipsla_monitoring_policy).
+        type: str
+      ipsla_monitoring_policy:
+        description:
+        - The name IPSLA Monitoring Policy to use for the member.
+        - This parameter can be used instead of O(members.ipsla_monitoring_policy_uuid).
+        type: str
+        aliases: [ ipsla_monitoring_policy_name ]
```
akinross, line 103: If we decide to input uuid for these nested inputs, should we then do it for all objects that have an uuid? think template in that case also has an uuid.

plugins/modules/ndo_ipsla_track_list.py
```diff
@@ -0,0 +1,490 @@
...
+        suboptions:
+          name:
+            description:
+            - The name of the BD or L3Out used as the scope for the member.
+            type: str
+            required: true
+          schema:
+            description:
+            - The name of the Schema associated with the BD scope.
+            - This parameter is only required when the O(members.scope_type) is V(bd).
```
akinross, line 129:
```suggestion
            - This parameter is only required when the O(members.scope_type=bd).
```

akinross (changes requested):

plugins/modules/ndo_ipsla_track_list.py
samiib: Do you mean template within the scope sub-option or the top level template?

The scope template is only required if the L3out or BD UUID isn't provided. I think having scope template UUID as an options might make the module a little more complex than it needs to be.

If template UUID is an option at the top level then we would have to make it consistent everywhere and split that into a separate task.

plugins/modules/ndo_ipsla_track_list.py
```diff
@@ -0,0 +1,490 @@
...
+                    weight=dict(type="int"),
+                ),
+                mutually_exclusive=[
+                    ("ipsla_monitoring_policy_uuid", "ipsla_monitoring_policy"),
+                    ("scope_uuid", "scope"),
+                ],
+                required_one_of=[
+                    ("ipsla_monitoring_policy_uuid", "ipsla_monitoring_policy"),
+                    ("scope_uuid", "scope"),
+                ],
```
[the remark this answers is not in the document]
samiib: This spec is basically enforcing two things:
1. Either `ipsla_monitoring_policy_uuid` or `ipsla_monitoring_policy` is required but not both at the same time.
2. Either `scope_uuid` or `scope` is required but not both at the same time.

I don't think its possible to enforce this rule for `ipsla_track_list_uuid/name` because the one case of updating the name requires providing both options. We already enforce a rule when that state is present or absent either `ipsla_track_list_uuid/name` is required.

plugins/modules/ndo_ipsla_track_list.py
```diff
@@ -0,0 +1,490 @@
...
+    valid_lower = 0
+    if list_type == "weight":
+        valid_upper = 255
+    for threshold_key, threshold_value in thresholds.items():
+        if threshold_value is not None and threshold_value not in range(valid_lower, valid_upper):
+            mso.fail_json(
+                msg="Invalid value provided for threshold_{0}: {1}; it must be in the range {2} - {3}".format(
+                    threshold_key, threshold_value, valid_lower, valid_upper
+                )
+            )
```
gmicol, line 360: Have you tested if the API returns an error if the `threshold_value` is not in range? If the API does return a proper error for every possible scenario where the `threshold_value` is out of range, I think it's not needed to add this validation segment

plugins/modules/ndo_ipsla_track_list.py
```diff
@@ -0,0 +1,490 @@
...
+
+    if thresholds["down"] is not None and thresholds["up"] is not None:
+        if thresholds["down"] > thresholds["up"]:
+            mso.fail_json(
+                msg="Invalid value provided for threshold_down: {0}; it must be less than or equal to threshold_up: {1}".format(
+                    thresholds["down"], thresholds["up"]
+                )
+            )
```
gmicol, line 368: Same question as before regarding the API's behavior in this scenario

gmicol (changes requested):

plugins/modules/ndo_ipsla_track_list.py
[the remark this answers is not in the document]
gmicol: Why not simply prioritize the UUID over the name by ignoring the name if UUID is provided? It's what you've been so far in the recent new module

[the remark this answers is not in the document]
samiib: Based on a user perspective I think its a better experience to be explicit that the two options do not apply together. This reduces any potential confusion and makes the expected behaviour clear.
I recommended a similar change in this PR -> https://github.com/CiscoDevNet/ansible-mso/pull/577#discussion_r1922401522

plugins/modules/ndo_ipsla_track_list.py
samiib: The API doesn't return a consistently nice error message when the thresholds are out of range for all scenarios. When the lower range is less than 0, eg -1, the message not quite acceptable.

`MSO Error 400: json: cannot unmarshal number -1 into Go struct field Template2.tenantPolicyTemplate of type uint32`

I'll keep this validation in the module so the threshold range error is unambiguous.

plugins/modules/ndo_ipsla_track_list.py
samiib: The API returns a clear error message in this case. Hence I have removed this extra validation.

plugins/modules/ndo_ipsla_track_list.py
akinross: Any of them that allow uuid input I mean. The original uuid input was only to change a name of the object. Later a decision was made to do it for more, but now I see we then do not do it for everything, thus my question.

Regarding complexity, I think the uuid as input anywhere besides the name change use-case already makes things more complex than it needs to be.

plugins/modules/ndo_ipsla_track_list.py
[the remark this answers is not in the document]
akinross: @samiib, I have argued the same as your comment in earlier discussions. Then it was decided we would prioritise the UUID over the name by ignoring the name if UUID is provided as Gaspard mentioned. This is also the reason for my question above.

plugins/modules/ndo_ipsla_track_list.py
samiib: The main reason to provide UUID as an option for any input, apart from the name change case, would be if there is a performance benefit. In this module using UUID instead of template/schema/name for `scope` is much quicker. This is because we would use the UUID directly instead of having to search for UUID from a potentially different template.

I think UUID for IPSLA Monitoring Policy would been useful if it could reference a policy from a different template. Initially I thought this was the case but its not and now I can remove `ipsla_monitoring_policy_uuid`.

plugins/modules/ndo_ipsla_track_list.py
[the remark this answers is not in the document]
samiib: Removed this spec. `scope_uuid` now takes precedence over `scope`.

plugins/modules/ndo_ipsla_track_list.py
samiib: Removed `ipsla_monitoring_policy_uuid`.

gmicol (dismissed):
LGTM

akinross (approved):
LGTM

sajagana (dismissed):
LGTM!

shrsr (approved):
LGTM

anvitha-jain (dismissed):
LGTM

plugins/modules/ndo_ipsla_track_list.py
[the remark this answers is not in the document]
samiib: As decided in stand-up. I have reverted back to what I originally thought would be beneficial and now the spec uses `mutually_exclusive` for `scope` and `scope_uuid`.

plugins/modules/ndo_ipsla_track_list.py
```diff
@@ -0,0 +1,468 @@
...
+    if ipsla_track_list or ipsla_track_list_uuid:
+        match = mso_template.get_object_by_key_value_pairs(
+            object_description,
+            existing_ipsla_track_lists,
+            [(KVPair("uuid", ipsla_track_list_uuid) if ipsla_track_list_uuid else KVPair("name", ipsla_track_list))],
+        )
+        if match:
+            mso.existing = mso.previous = copy.deepcopy(match.details)
+    else:
+        mso.existing = mso.previous = existing_ipsla_track_lists
```
sajagana, line 372: Insert the `ipslaMonitoringRef` and `scope` object name in the `current, previous` output portion when we do the `create, update, query, delete`.

plugins/modules/ndo_ipsla_track_list.py
```diff
@@ -0,0 +1,468 @@
...
+            object_description,
+            ipsla_track_lists,
+            [(KVPair("uuid", ipsla_track_list_uuid) if ipsla_track_list_uuid else KVPair("name", ipsla_track_list))],
+        )
+        if match:
+            mso.existing = match.details
+        else:
+            mso.existing = {}
+    elif module.check_mode and state != "query":
+        mso.existing = mso.proposed if state == "present" else {}
```
sajagana, line 411: Insert the `ipslaMonitoringRef` and `scope` object name in the `current, previous` output portion when we do the `create, update, query, delete`.

plugins/modules/ndo_ipsla_track_list.py
```diff
@@ -0,0 +1,468 @@
...
+            type=dict(type="str", choices=["percentage", "weight"]),
+            threshold_up=dict(type="int", aliases=["up"]),
+            threshold_down=dict(type="int", aliases=["down"]),
+            members=dict(
+                type="list",
+                elements="dict",
+                mutually_exclusive=[("scope", "scope_uuid")],
+                options=dict(
+                    destination_ip=dict(type="str", aliases=["ip"], required=True),
+                    ipsla_monitoring_policy=dict(type="str", aliases=["ipsla_monitoring_policy_name"]),
```
sajagana, line 293: The `ipsla_monitoring_policy_uuid` input option is missing.

plugins/modules/ndo_ipsla_track_list.py
```diff
@@ -0,0 +1,468 @@
...
+    append_update_ops_data,
+)
+import copy
+
+
+def main():
+    argument_spec = mso_argument_spec()
+    argument_spec.update(
+        dict(
+            template=dict(type="str", required=True),
```
sajagana, line 280: The `template_id` input option is missing.

sajagana (changes requested):

plugins/module_utils/schema.py
```diff
@@ -24,6 +25,15 @@ def __init__(self, mso_module, schema_name, template_name=None, site_name=None):
         if site_name and template_name:
             self.set_site(template_name, site_name)
 
+    @classmethod
+    def with_template_id(cls, mso_module, schema_name, template_name=None, template_id=None):
```
akinross, line 29: why add this as class method? since it has default value of None can't it be added to the constructor?

plugins/modules/ndo_ipsla_track_list.py
```diff
@@ -0,0 +1,548 @@
...
+                    ),
+                    ipsla_monitoring_policy_uuid=dict(type="str"),
+                    scope_uuid=dict(type="str"),
+                    scope=dict(
+                        type="dict",
+                        options=dict(
+                            name=dict(type="str", required=True),
+                            template=dict(type="str"),
+                            template_id=dict(type="str"),
+                            schema=dict(type="str"),
```
akinross, line 339: since we are doing id now everywhere should we then also do this for schema?

plugins/modules/ndo_ipsla_track_list.py
```diff
@@ -0,0 +1,548 @@
...
+    members = module.params.get("members")
+    state = module.params.get("state")
+
+    # Validate
+    valid_upper = 100
+    valid_lower = 0
+    if list_type == "weight":
+        valid_upper = 255
+    for threshold_key, threshold_value in thresholds.items():
+        if threshold_value is not None and threshold_value not in range(valid_lower, valid_upper):
```
akinross, line 393: should we append all the potential errors (up and down) in threshold to fail in single error?

akinross (changes requested):

plugins/modules/ndo_ipsla_track_list.py
samiib: There are only 2 thresholds in the dictionary, up and down. We don't usually append all validation errors into one message.
If the validation was in a list of items it would be useful but I think because its only two options this should be okay.

akinross (dismissed):
LGTM

shrsr (dismissed):
LGTM

gmicol (approved):
LGTM

anvitha-jain (dismissed):
LGTM

gmicol (dismissed):
LGTM

sajagana (approved):
LGTM!

lhercot (changes requested):
Need a rebase

gmicol (approved):
LGTM

plugins/module_utils/template.py
```diff
@@ -320,26 +346,34 @@ def get_template_object_name_by_uuid(self, object_type, uuid, fail_module=True):
...
+        :param object_type: The type of the object to retrieve -> Str
+        :param uuid: The UUID of the object to retrieve -> Str
+        :param use_cache: Use the cached result of the templates/objects API for the UUID -> Bool
         :return: Dict | None: The processed result which could be:
             When the UUID is existing, returns object -> Dict
             When the UUID is not existing -> None
         """
-        response_object = self.mso.request("templates/objects?type={0}&uuid={1}".format(object_type, uuid), "GET")
+        response_object = None
+        if use_cache and uuid in self.template_objects_cache.keys():
```
shrsr, line 360: Can we just do away with use_cache?
```
if uuid in self.template_objects_cache.keys():
    response_object = self.template_objects_cache[uuid]
```

plugins/modules/ndo_ipsla_track_list.py
```diff
@@ -0,0 +1,566 @@
...
+        "schema": "scopeSchemaName",
+        "schemaId": "scopeSchemaId",
+        "type": "bd",
+    }
+    ipsla_ref = {
+        "name": "ipslaMonitoringPolicyName",
+        "reference": "ipslaMonitoringRef",
+        "type": "ipslaMonitoringPolicy",
+    }
+    mso_template.clear_template_objects_cache()
```
shrsr, line 552: Are we using this here because the object related to uuid can be modified?

shrsr (changes requested):

plugins/module_utils/template.py
samiib: I would need to move `if use_cache:` down to line 364 where it adds the object to the cache instead.
Or do you think we can remove `use_cache` and always cache? I was thinking of making it optional at first so it doesn't create any unexpected side effects in other modules.

plugins/modules/ndo_ipsla_track_list.py
samiib: There is the potential that the referenced object details could change between the initial match and after an update. However unlikely, this just makes sure the latest objects are used each time. The cache just helps when the same objects are referenced in multiple members.

akinross (approved):
LGTM

shrsr (approved):
LGTM

anvitha-jain (approved):
LGTM

sajagana (approved):
LGTM!

lhercot (approved):
LGTM

PR #610, opened 2025-02-10
[minor_change] Added support for Ansible 2.18 and dropped support for 2.15 (DCNE-250)

samiib:


gmicol (approved):
LGTM

shrsr (changes requested):
Why are additional changed being made to the modules apart from changing the Ansible support?
Example-
```
if mso.existing and domain_path:
```

samiib:
> Why are additional changed being made to the modules apart from changing the Ansible support? Example-
> 
> ```
> if mso.existing and domain_path:
> ```

@shrsr These are related to the additional sanity issues that get reported in 2.18. 
There are a few cases of variables being declared within conditional statements but being used outside it later. This created an unhandled unbound variable scenario in many modules. I thought it would be better to handle the case safely while fixing the sanity issues. Functionally there should be no change to these modules.

shrsr (approved):
LGTM

.github/workflows/ansible-test.yml
```diff
@@ -7,15 +7,15 @@ on:
     # * is a special character in YAML so you have to quote this string
     - cron: '0 7 * * *'
 env:
-  python_version: '3.10'
+  python_version: '3.11'
```
sajagana, line 10: Can we use 3.12 or 3.13?

plugins/modules/ndo_interface_setting.py
```diff
@@ -559,7 +560,7 @@ def main():
         if uuid and not mso.existing:
             mso.fail_json(msg="{0} with the UUID: '{1}' not found".format(object_description, uuid))
 
-        if mso.existing:
+        if mso.existing and match:
```
sajagana, line 563: It should be `match` or `interface_path`?

You can remove the `match` from the check when you check `mso.existing`.

plugins/modules/ndo_physical_interface.py
```diff
@@ -332,7 +333,7 @@ def main():
         if interface_descriptions:
             mso_values["interfaceDescriptions"] = format_interface_descriptions(mso, interface_descriptions, "")
 
-        if mso.existing:
+        if mso.existing and match:
```
sajagana, line 336: It should be `match` or `physical_interface_attrs_path`?

plugins/modules/ndo_port_channel_interface.py
```diff
@@ -336,7 +338,7 @@ def main():
             description=description,
         )
 
-        if mso.existing:
+        if mso.existing and match:
```
sajagana, line 341: Same question for all other modules.

sajagana (changes requested):

.github/workflows/ansible-test.yml
samiib: We have been using the minimum required version of python here for now. It will get upgraded when a new version of Ansible stops supporting the Python version. I would be nice to use the latest python but it will be hard for us to keep up with the releases.

plugins/modules/ndo_interface_setting.py
samiib: I mentioned in a previous comment that I added the extra conditional so we can safely handle unbound variable scenarios. The parts below this conditional all use `match` and the `_path` variables. However, the `_path` variables are only set when there is a `match`. Its possible for `mso.existing` to be set without a match when name/UUID isn't provided, even though that doesn't usually happen. 

I could check all 3 variables `if mso.existing and match and interface_path`. Although, I think existing and `match` is enough here.

akinross (approved):
LGTM

sajagana (approved):
LGTM!

anvitha-jain (approved):
LGTM

lhercot (approved):
LGTM

PR #611, opened 2025-02-13
Added ndo_l3out_bgp_peer module to manage L3Out Node/Interface Group Policy - BGP Peer objects (DCNE-182)

sajagana:


plugins/modules/ndo_l3out_bgp_peer.py
```diff
@@ -0,0 +1,735 @@
...
+        if peer_prefix:
+            peer_prefix_objects = mso.query_objs(generate_api_endpoint("templates/objects", **{"type": "bgpPeerPrefixPol", "tenant-id": tenant_id}))
+            peer_prefix_uuid = get_peer_prefix_object(mso_template, peer_prefix_uuid, peer_prefix, peer_prefix_objects).details.get("uuid")
+
+        mso_values = dict(
+            peerAddressV4=ipv4_addr,
+            peerAddressV6=ipv6_addr,
+            peerAsn=remote_asn,
+            adminState=admin_state,
+            authEnabled=True if auth_password else None,
```
akinross, line 507: so false is not a valid option?

plugins/modules/ndo_l3out_bgp_peer.py
```diff
@@ -0,0 +1,735 @@
...
+            mso_values["exportRouteMapRef"] = mso_template.get_route_map(
+                "export_route_map",
+                tenant_id,
+                tenant_name,
+                export_route_map,
+                route_map_objects,
+            ).get("uuid", None)
+
+        # BGP Controls
+        bgp_controls_key_mapping = {
```
akinross, line 539: shoudl this be defined in constants? same applies to the other mappings below

plugins/modules/ndo_l3out_bgp_peer.py
```diff
@@ -0,0 +1,735 @@
...
+            mso_values["peerControls"] = peer_controls_map if peer_controls_map is not None else None
+            mso_values["addressTypeControls"] = address_families_map if address_families_map is not None else None
+            mso_values["privateASControls"] = private_as_controls_map if private_as_controls_map is not None else None
+            cleaned_mso_values = remove_none_values(mso_values)
+            mso.sanitize(cleaned_mso_values)
+            ops.append(dict(op="add", path=bgp_peer_path, value=cleaned_mso_values))
+
+        elif mso.existing:
+            proposed_payload = copy.deepcopy(mso.existing)
+            mso_values_remove = list()
```
akinross, line 584: what is the use of this values remove list?

plugins/modules/ndo_l3out_bgp_peer.py
```diff
@@ -0,0 +1,735 @@
...
+        if private_as_controls_map and (private_as_controls_map.get("removeAll") or private_as_controls_map.get("replaceWithLocalAS")):
+            private_as_controls_map["removeExclusive"] = True
+
+        if not mso.existing:
+            mso_values["bgpControls"] = bgp_controls_map if bgp_controls_map is not None else None
+            mso_values["peerControls"] = peer_controls_map if peer_controls_map is not None else None
+            mso_values["addressTypeControls"] = address_families_map if address_families_map is not None else None
+            mso_values["privateASControls"] = private_as_controls_map if private_as_controls_map is not None else None
+            cleaned_mso_values = remove_none_values(mso_values)
+            mso.sanitize(cleaned_mso_values)
```
akinross, line 579: is sanitize not already removing the none values?

akinross (changes requested):

plugins/modules/ndo_l3out_bgp_peer.py
```diff
@@ -0,0 +1,735 @@
...
+author:
+- Sabari Jaganathan (@sajagana)
+options:
+  template:
+    description:
+    - The name of the L3Out template.
+    type: str
+    aliases: [ l3out_template ]
+    required: true
+  l3out:
```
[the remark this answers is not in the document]
sajagana: No, this is not a primary attribute for this module.

plugins/modules/ndo_l3out_bgp_peer.py
sajagana: False is valid option, when the `auth_password` is None, `authEnabled` sets to `false` by default.

plugins/modules/ndo_l3out_bgp_peer.py
sajagana: `mso_values_remove` is used to store the attributes which need to be cleared from the existing object.

anvitha-jain (dismissed):
LGTM

shrsr (changes requested):

plugins/modules/ndo_l3out_bgp_peer.py
```diff
@@ -0,0 +1,800 @@
...
+    type: str
+    choices: [ none, no_prepend, dual_as, replace_as ]
+  local_asn:
+    description:
+    - The local autonomous system number (ASN) of the L3Out BGP Peer.
+    - The value must be between 1 and 4294967295.
+    type: int
+  import_route_map:
+    description:
+    - The name of the import route map.
```
samiib, line 107:
```suggestion
    - The name of the import route map.
    - This parameter can be used instead of O(import_route_map_uuid).
```

plugins/modules/ndo_l3out_bgp_peer.py
```diff
@@ -0,0 +1,800 @@
...
+    - Providing an empty string will remove the O(import_route_map="") from the L3Out BGP Peer.
+    type: str
+  import_route_map_uuid:
+    description:
+    - The UUID of the import route map.
```
samiib, line 112:
```suggestion
    - The UUID of the import route map.
    - This parameter can be used instead of O(import_route_map).
```

plugins/modules/ndo_l3out_bgp_peer.py
```diff
@@ -0,0 +1,800 @@
...
+    description:
+    - The configuration for local ASN of the L3Out BGP Peer.
+    type: str
+    choices: [ none, no_prepend, dual_as, replace_as ]
+  local_asn:
+    description:
+    - The local autonomous system number (ASN) of the L3Out BGP Peer.
+    - The value must be between 1 and 4294967295.
+    type: int
+  import_route_map:
```
samiib, line 105: Should the `route_map` and `peer_prefix` parameters be a dictionary with `name` and `template` options? How do we specify which template the objects are from?

plugins/modules/ndo_l3out_bgp_peer.py
```diff
@@ -0,0 +1,800 @@
...
+    - Providing an empty string will remove the O(import_route_map="") from the L3Out BGP Peer.
+    type: str
+  import_route_map_uuid:
+    description:
+    - The UUID of the import route map.
+    - Providing an empty string will remove the O(import_route_map_uuid="") from the L3Out BGP Peer.
+    type: str
+  export_route_map:
+    description:
+    - The name of the export route map.
```
samiib, line 117:
```suggestion
    - The name of the export route map.
    - This parameter can be used instead of O(export_route_map_uuid).
```

plugins/modules/ndo_l3out_bgp_peer.py
```diff
@@ -0,0 +1,800 @@
...
+    - Providing an empty string will remove the O(export_route_map="") from the L3Out BGP Peer.
+    type: str
+  export_route_map_uuid:
+    description:
+    - The UUID of the export route map.
```
samiib, line 122:
```suggestion
    - The UUID of the export route map.
    - This parameter can be used instead of O(export_route_map).
```

plugins/modules/ndo_l3out_bgp_peer.py
```diff
@@ -0,0 +1,800 @@
...
+    - Providing an empty string will remove the O(export_route_map_uuid="") from the L3Out BGP Peer.
+    type: str
+  peer_prefix:
+    description:
+    - The name of the peer prefix.
```
samiib, line 127:
```suggestion
    - The name of the peer prefix.
    - This parameter can be used instead of O(peer_prefix_uuid).
```

plugins/modules/ndo_l3out_bgp_peer.py
```diff
@@ -0,0 +1,800 @@
...
+    - Providing an empty string will remove the O(peer_prefix="") from the L3Out BGP Peer.
+    type: str
+  peer_prefix_uuid:
+    description:
+    - The UUID of the peer prefix.
```
samiib, line 132:
```suggestion
    - The UUID of the peer prefix.
    - This parameter can be used instead of O(peer_prefix).
```

samiib (changes requested):

plugins/modules/ndo_l3out_bgp_peer.py
sajagana: The `route_map` objects are unique within the tenant level, so we can skip the template check for the `route_map` objects.
The `peer_prefix` also unique within the tenant level, but it allows to import common tenant `peer_prefix` value. Let me make changes for both route map and peer prefix.

sajagana: The `route_map` objects are unique within the tenant level, so we can skip the template check for the `route_map` objects.
The `peer_prefix` also unique within the tenant level, but it allows to import common tenant `peer_prefix` value. Let me make changes for both route map and peer prefix.

plugins/modules/ndo_l3out_bgp_peer.py
```diff
@@ -0,0 +1,800 @@
...
+
+    if state == "present":
+        peer_prefix_uuid = None
+
+        if peer_prefix_identifier.get("uuid") is not None:
+            peer_prefix_uuid = peer_prefix_identifier.get("uuid")
+        elif peer_prefix_identifier.get("name") is not None and peer_prefix_identifier.get("name") != "":
+            peer_prefix_uuid = mso_template.get_object_by_key_value_pairs(
+                "BGP Peer Prefix Policy",
+                mso.query_objs(generate_api_endpoint("templates/objects", **{"type": "bgpPeerPrefixPol", "tenant-id": tenant_id})),
```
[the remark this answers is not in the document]
sajagana: The template objects are unique within the tenant level, so we cannot have the same name object with two different templates. Let me recheck the logic once again.

plugins/modules/ndo_l3out_bgp_peer.py
akinross: see choices made in this discussion where it is a map with name, https://github.com/CiscoDevNet/ansible-mso/pull/577/files#r1970341772. perhaps we should follow same here?

plugins/modules/ndo_l3out_bgp_peer.py
```diff
@@ -0,0 +1,800 @@
...
+            ("l3out", "l3out_uuid"),
+            ("import_route_map", "import_route_map_uuid"),
+            ("export_route_map", "export_route_map_uuid"),
+            ("peer_prefix", "peer_prefix_uuid"),
+        ],
+    )
+
+    mso = MSOModule(module)
+
+    template_identifier = get_object_identifier(module.params.get("template_id"), module.params.get("template"))
```
akinross, line 489: what is the use case for template_identifier instead of just storing in two variables?

plugins/modules/ndo_l3out_bgp_peer.py
```diff
@@ -0,0 +1,800 @@
...
+    module = AnsibleModule(
+        argument_spec=argument_spec,
+        supports_check_mode=True,
+        required_if=[
+            ["state", "present", ["template", "template_id"], True],
+            ["state", "query", ["template", "template_id"], True],
+            ["state", "absent", ["template", "template_id"], True],
+            ["state", "present", ["l3out", "l3out_uuid"], True],
+            ["state", "query", ["l3out", "l3out_uuid"], True],
+            ["state", "absent", ["l3out", "l3out_uuid"], True],
```
akinross, line 474: Is there a reason to do this over required_one_of and mutually_exclusive? I understand the required_if works but from code reading it might be easier to understand.

In ansible docs https://docs.ansible.com/ansible/latest/dev_guide/developing_program_flow_modules.html#argument-spec-dependencies the following is mentioned: n this example, at least one of path and content must be specified. If none are specified, execution will fail. Specifying both is explicitly allowed; to prevent this, combine required_one_of with mutually_exclusive.

plugins/modules/ndo_l3out_bgp_peer.py
```diff
@@ -0,0 +1,800 @@
...
+            ).details.get("uuid")
+
+        route_map_objects = (
+            mso.query_objs(generate_api_endpoint("templates/objects", **{"type": "routeMap", "tenant-id": tenant_id}))
+            if import_route_map_identifier.get("name") or export_route_map_identifier.get("name")
+            else []
+        )
+
+        import_route_map_uuid = None
+        export_route_map_uuid = None
```
akinross, line 559: I am not sure why this is done here, why not do it at the beginning where we do module.params.get? Like first you create a dict, but then later you are still can declare the individual variables? Is it not easier to write the uuid to the variable module.params.get and then when this value is None set this variable with mso_template.get_route_map() method?

akinross (changes requested):

tests/integration/targets/ndo_l3out_bgp_peer/tasks/main.yml
```diff
@@ -0,0 +1,1034 @@
...
+          - add_bgp_peer_with_default_ctrl.current.peerAddressV4 == "3.3.3.3"
+          - add_bgp_peer_with_default_ctrl.current.peerAddressV6 == "1::9/16"
+          - add_bgp_peer_with_default_ctrl.previous == {}
+
+    # UPDATE
+    - name: Clear route map and peer prefix config from the L3Out BGP Peer using UUID with empty string (check_mode)
+      cisco.mso.ndo_l3out_bgp_peer:
+        <<: *mso_info
+        template_id: "{{ query_l3out_policy_template.current.templateId }}"
+        l3out: "l3out_1"
```
akinross, line 412: is there a reason to skip testing of l3out_uuid?

akinross (changes requested):

plugins/modules/ndo_l3out_bgp_peer.py
```diff
@@ -0,0 +1,800 @@
...
+
+            if ipv6_addr == "" and "peerAddressV6" in proposed_payload:
+                mso_values_remove.append("peerAddressV6")
+                mso_values.pop("peerAddressV6", None)
+
+            if remote_asn == 0 and "peerAsn" in proposed_payload:
+                mso_values_remove.append("peerAsn")
+                mso_values.pop("peerAsn", None)
+
+            if weight == 0 and "weight" in proposed_payload:
```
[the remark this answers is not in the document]
sajagana: For now, `0` will clear the configuration.

plugins/modules/ndo_l3out_bgp_peer.py
```diff
@@ -0,0 +1,800 @@
...
+
+            if ipv4_addr == "" and "peerAddressV4" in proposed_payload:
+                mso_values_remove.append("peerAddressV4")
+                mso_values.pop("peerAddressV4", None)
+
+            if ipv6_addr == "" and "peerAddressV6" in proposed_payload:
+                mso_values_remove.append("peerAddressV6")
+                mso_values.pop("peerAddressV6", None)
+
+            if remote_asn == 0 and "peerAsn" in proposed_payload:
```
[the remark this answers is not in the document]
sajagana: For now, `0` will clear the configuration.

plugins/modules/ndo_l3out_bgp_peer.py
```diff
@@ -0,0 +1,880 @@
...
+    username: admin
+    password: SomeSecretPassword
+    template: l3out_template
+    l3out: l3out_1
+    node_group: node_group_policy_1
+    ipv6_address: "1::8/16"
+    state: query
+  register: query_with_ipv6
+
+- name: Remove an L3Out BGP Peer with IPv4 and IPv6 addresses
```
shrsr, line 412: Can we add an example for query_all?

shrsr (changes requested):

plugins/modules/ndo_l3out_bgp_peer.py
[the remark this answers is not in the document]
shrsr: Why did we decide this?

plugins/module_utils/mso.py
```diff
@@ -1302,6 +1351,20 @@ def sanitize(self, updates, collate=False, required=None, unwanted=None):
...
+                if not collate and updates.get(key) is not None:
+                    cleaned_value = self.remove_none_values(updates.get(key))
+                    self.sent[key] = (
+                        cleaned_value
+                        if cleaned_value
+                        else (
+                            {}
+                            if isinstance(updates.get(key), dict)
+                            else ([] if isinstance(updates.get(key), list) else (cleaned_value if cleaned_value is not None else None))
+                        )
```
shrsr, line 1364: I'm curious:
Would we ever encounter this scenario here and should we remove the key child completely from sent for the below case?

before -
    "parent": {
      "child": {
        "multicast": None,
      }

after -
"parent": {
      "child": {
      }

plugins/module_utils/mso.py
```diff
@@ -1302,6 +1351,20 @@ def sanitize(self, updates, collate=False, required=None, unwanted=None):
...
+                # Remove nested unspecified values
+                if not collate and updates.get(key) is not None:
+                    cleaned_value = self.remove_none_values(updates.get(key))
+                    self.sent[key] = (
+                        cleaned_value
+                        if cleaned_value
+                        else (
+                            {}
+                            if isinstance(updates.get(key), dict)
+                            else ([] if isinstance(updates.get(key), list) else (cleaned_value if cleaned_value is not None else None))
```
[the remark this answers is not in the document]
sajagana: Yes, when the `cleaned_value` is `False`, I want to return `False` instead of None. Without the below check I could not return `False`.

`(cleaned_value if cleaned_value is not None else None)`

plugins/module_utils/mso.py
sajagana: Yes, when the `child` contains only `None` value attributes, we remove those attributes. The final output will be `child={}`.

plugins/modules/ndo_l3out_bgp_peer.py
sajagana: Done

plugins/modules/ndo_l3out_bgp_peer.py
```diff
@@ -0,0 +1,880 @@
...
+    choices: [ absent, query, present ]
+    default: query
+notes:
+- The O(template) must exist before using this module in your playbook.
+  Use M(cisco.mso.ndo_template) to create the L3Out template.
+- The O(l3out) must exist before using this module in your playbook.
+  Use M(cisco.mso.ndo_l3out_template) to create the L3Out.
+- The O(node_group) must exist before using this module in your playbook.
+  Use M(cisco.mso.ndo_l3out_node_group_policy) to create the L3Out Node Group Policy.
+- The O(peer_prefix) must exist before using this module in your playbook.
```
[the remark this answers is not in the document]
sajagana: Done

plugins/module_utils/mso.py
shrsr: Should we delete the key child then and in turn parent because they become unspecified at this point? This is just hypothetical..

shrsr: Can we move around the code and do this? I'm not sure what the behaviour should be but this is a suggestion..

```
# Remove nested unspecified values
if not collate and updates.get(key) is not None:
    cleaned_value = self.remove_none_values(updates[key])
    
    if cleaned_value is not None:
        self.sent[key] = cleaned_value
    else:
        updates[key] = None

# Remove unspecified values
if not collate and updates.get(key) is None:
    if key in self.existing:
        del self.sent[key]
    continue

# Remove identical values
elif not collate and updates.get(key) == self.existing.get(key):
    del self.sent[key]
    continue
```

shrsr: or just make this change?

```
# Remove nested unspecified values
if not collate and updates.get(key) is not None:
    cleaned_value = self.remove_none_values(updates[key])
    
    if cleaned_value is not None:
        self.sent[key] = cleaned_value
    else:
        del self.sent[key]
    continue
```

plugins/modules/ndo_l3out_bgp_peer.py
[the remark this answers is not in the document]
sajagana: To prevent a type mismatch, I'm using `0` to clear the `weight` attribute, not an empty string.

Allowing both empty strings and integers for the same attribute creates unnecessary conflicts for module consumers.

plugins/module_utils/mso.py
sajagana: Since it has the significant impact on existing mso modules, removed the `remove_none_values` function from the mso.py.

shrsr (dismissed):
Apart from '0' being used for deletion of certain integer values everything else LGTM.

plugins/modules/ndo_l3out_bgp_peer.py
```diff
@@ -0,0 +1,934 @@
...
+
+        if local_asn not in ["", None]:
+            mso_values["localAsn"] = int(local_asn)
+
+        if not mso.existing:
+            # BGP Controls
+            if bgp_controls:
+                bgp_controls_dict = dict()
+
+                if bgp_controls["allow_self_as"]:
```
akinross, line 692: same value assignment question as above applies to the other dict assignments in your code

plugins/modules/ndo_l3out_bgp_peer.py
```diff
@@ -0,0 +1,934 @@
...
+
+def initiate_l3out_bgp_peer_mso_templates(mso, template_type, template_name, template_id):
+    key = (template_name, template_id)
+
+    # Check if the combination is unique and create an instance if needed
+    if key not in l3out_bgp_peer_instance_cache:
+        l3out_bgp_peer_instance_cache[key] = MSOTemplate(mso, template_type, template_name, template_id)
+
+    # Return the instance for this key
+    return l3out_bgp_peer_instance_cache[key]
```
akinross, line 930: Should we define a generic functionality for template caching? I think I did caching before in some older modules for schema caching and @samiib also introduced some form of caching a while ago.

akinross (changes requested):

plugins/modules/ndo_l3out_bgp_peer.py
```diff
@@ -0,0 +1,934 @@
...
+        bgp_peer_path = "/l3outTemplate/l3outs/{0}/nodeGroups/{1}/bgpPeers/{2}".format(
+            l3out_object.index, node_group_object.index, bgp_peer_objects.index if bgp_peer_objects else "-"
+        )
+
+    ops = []
+
+    if state == "present":
+        peer_prefix_is_empty = check_if_all_elements_are_none(list(peer_prefix.values())) if peer_prefix else True  # peer_prefix is None or empty dict => True
+        import_route_map_is_empty = check_if_all_elements_are_none(list(import_route_map.values())) if import_route_map else True
+        export_route_map_is_empty = check_if_all_elements_are_none(list(export_route_map.values())) if export_route_map else True
```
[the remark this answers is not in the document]
sajagana: Done!

plugins/modules/ndo_l3out_bgp_peer.py
sajagana: Handling nested `None` values is not possible with current mso sanitize function.

plugins/modules/ndo_l3out_bgp_peer.py
sajagana: Done!

plugins/module_utils/template.py
```diff
@@ -429,3 +429,30 @@ def update_config_with_template_and_references(self, config_data, reference_coll
     def check_template_when_name_is_provided(self, parameter):
         if parameter and parameter.get("name") and not (parameter.get("template") or parameter.get("template_id")):
             self.mso.fail_json(msg="Either 'template' or 'template_id' associated with '{}' must be provided".format(parameter.get("name")))
+
+
+class MSOTemplateCache:
+    def __init__(self):
+        self.cache = {}
+        self.template_property_cache = {}
```
akinross, line 437: not sure I understand why this is needed?

akinross (changes requested):

plugins/module_utils/template.py
sajagana: Done!

akinross (approved):
LGTM

shrsr (approved):
LGTM

anvitha-jain (approved):

gmicol (approved):
LGTM

plugins/module_utils/template.py
```diff
@@ -440,3 +440,17 @@ def get_route_map_policy_for_multicast_uuid(self, route_map_policy_for_multicast
         kv_list = [KVPair("name", route_map_policy_for_multicast_name)]
         match = self.get_object_by_key_value_pairs("Route Map Policy for Multicast", existing_route_map_policies, kv_list, fail_module=True)
         return match.details.get("uuid")
+
+
+class MSOTemplateCache:
```
shrsr, line 445: If it possible to move self.cache = {} to init function or constructor of class MSOTemplate itself?

plugins/module_utils/template.py
```diff
@@ -440,3 +440,17 @@ def get_route_map_policy_for_multicast_uuid(self, route_map_policy_for_multicast
...
+    def __init__(self):
+        self.cache = {}
+
+    def get_template(self, mso, template_type, template_name, template_id):
```
shrsr, line 449: Can we move this to MSOTemplate?

```
def get_template(self, template_type, template_name, template_id):
        if template_id in self.cache:
            return self.cache[template_id]

        new_template = MSOTemplate(self.mso, template_type, template_name, template_id)
        self.cache[new_template.template_id] = new_template
        self.cache[(new_template.template_name, new_template.template_type)] = new_template
        return new_template
```

plugins/modules/ndo_l3out_bgp_peer.py
```diff
@@ -0,0 +1,911 @@
...
+        )
+
+    ops = []
+
+    if state == "present":
+        peer_prefix_is_empty = check_if_all_elements_are_none(peer_prefix.values()) if peer_prefix else True  # peer_prefix is None or empty dict => True
+        import_route_map_is_empty = check_if_all_elements_are_none(import_route_map.values()) if import_route_map else True
+        export_route_map_is_empty = check_if_all_elements_are_none(export_route_map.values()) if export_route_map else True
+
+        mso_template_cache = MSOTemplateCache()
```
shrsr, line 643: We can do away with this if cache is moved to MSOTemplate itself and use mso_template object instantiated above?

shrsr (changes requested):

plugins/module_utils/template.py
sajagana: Done

plugins/module_utils/template.py
sajagana: Done

plugins/modules/ndo_l3out_bgp_peer.py
sajagana: Done

akinross (dismissed):
LGTM

akinross (approved):
LGTM

shrsr (dismissed):
LGTM

gmicol (dismissed):
LGTM

anvitha-jain (dismissed):

plugins/module_utils/template.py
```diff
@@ -440,3 +441,12 @@ def get_route_map_policy_for_multicast_uuid(self, route_map_policy_for_multicast
...
         match = self.get_object_by_key_value_pairs("Route Map Policy for Multicast", existing_route_map_policies, kv_list, fail_module=True)
         return match.details.get("uuid")
+
+    def get_template(self, template_type, template_name, template_id):
+        if template_id in self.cache:
+            return self.cache[template_id]
+
+        new_template = MSOTemplate(self.mso, template_type, template_name, template_id)
+        self.cache[new_template.template_id] = new_template
+        self.cache[(new_template.template_name, new_template.template_type)] = new_template
```
samiib, line 451: Should the name/type tuple key be checked above if template_id isn't provided?

sajagana: Done!

shrsr (approved):
LGTM

gmicol (approved):
LGTM

plugins/module_utils/mso.py
```diff
@@ -280,6 +280,24 @@ def ndo_bfd_multi_hop_settings_spec():
...
+            template_id=dict(type="str"),
+        ),
+        required_by={
+            "template": "name",
+            "template_id": "name",
+        },
+        mutually_exclusive=[
+            ["template", "template_id"],
+        ],
+    )
```
gmicol, line 298: Shouldn't we allow the option to add aliases to the name of the object in the spec:
```suggestion
def ndo_template_object_spec(aliases=None):
    return dict(
        type="dict",
        options=dict(
            name=dict(type="str", aliases=aliases) if aliases else dict(type="str"),
            template=dict(type="str"),
            template_id=dict(type="str"),
        ),
        required_by={
            "template": "name",
            "template_id": "name",
        },
        mutually_exclusive=[
            ["template", "template_id"],
        ],
    )
```

gmicol (changes requested):

plugins/module_utils/mso.py
sajagana: Done

shrsr (approved):
LGTM

akinross (approved):
LGTM

anvitha-jain (dismissed):
LGTM

samiib (approved):
LGTM

samiib (approved):
LGTM

shrsr (approved):
LGTM

gmicol (approved):
LGTM

anvitha-jain (approved):
LGTM

lhercot (approved):
LGTM

PR #614, opened 2025-02-24
[minor_change] Added new module for Node Profile. (DCNE-139)

### LINKED ISSUE #468: Enhancement: New module for Node Profile in Fabric Resource Policies Template (DCNE-139)
### Community Note

### Description

Create new module for Node Profile in Fabric Resource Policies Template ( created via ndo_template module )

* mso_XXXXX

* ND 3.1.1 / NDO 4.3.1

### Collection versions

* cisco.mso x.x.x

### References

* #0000

samiib:
Fixes #468

Also adds support for node ID ranges in ndo_physical_interface.

### DIFF
--- a/plugins/module_utils/mso.py
+++ b/plugins/module_utils/mso.py
@@ -1553,6 +1553,21 @@ class MSOModule(object):
         elif empty_attributes:
             self.module.fail_json(msg="When the '{0}' is '{1}', the {2} attributes must be set".format(attr_name, attr_value, empty_attributes))
 
+    def get_template_object_by_uuid(self, object_type, uuid, fail_module=True):
+        """
+        Retrieve a specific object type in the MSO template using its UUID.
+        :param object_type: The type of the object to retrieve the name for -> Str
+        :param uuid: The UUID of the object to retrieve the name for -> Str
+        :return: Dict | None: The processed result which could be:
+            When the UUID is existing, returns object -> Dict
+            When the UUID is not existing -> None
+        """
+        response_object = self.request("templates/objects?type={0}&uuid={1}".format(object_type, uuid), "GET")
+        if not response_object and fail_module:
+            msg = "Provided {0} with UUID of '{1}' not found.".format(object_type, uuid)
+            self.fail_json(msg=msg)
+        return response_object
+
     # Temporarily introduced method to handle nd specific query without introducing a dependency on the nd collection in code
     # Copied method from the nd collection: https://github.com/CiscoDevNet/ansible-nd/blob/master/plugins/module_utils/nd.py#L221
     # TODO: Refactor the code for bundled nd collection
--- a/plugins/module_utils/mso.py
+++ b/plugins/module_utils/mso.py
@@ -1553,21 +1553,6 @@ class MSOModule(object):
         elif empty_attributes:
             self.module.fail_json(msg="When the '{0}' is '{1}', the {2} attributes must be set".format(attr_name, attr_value, empty_attributes))
 
-    def get_template_object_by_uuid(self, object_type, uuid, fail_module=True):
-        """
-        Retrieve a specific object type in the MSO template using its UUID.
-        :param object_type: The type of the object to retrieve the name for -> Str
-        :param uuid: The UUID of the object to retrieve the name for -> Str
-        :return: Dict | None: The processed result which could be:
-            When the UUID is existing, returns object -> Dict
-            When the UUID is not existing -> None
-        """
-        response_object = self.request("templates/objects?type={0}&uuid={1}".format(object_type, uuid), "GET")
-        if not response_object and fail_module:
-            msg = "Provided {0} with UUID of '{1}' not found.".format(object_type, uuid)
-            self.fail_json(msg=msg)
-        return response_object
-
     # Temporarily introduced method to handle nd specific query without introducing a dependency on the nd collection in code
     # Copied method from the nd collection: https://github.com/CiscoDevNet/ansible-nd/blob/master/plugins/module_utils/nd.py#L221
     # TODO: Refactor the code for bundled nd collection
--- a/plugins/module_utils/template.py
+++ b/plugins/module_utils/template.py
@@ -95,6 +95,13 @@ class MSOTemplate:
         if self.template.get("templateType") != template_type:
             self.mso.fail_json(msg="Template type must be '{0}'.".format(template_type))
 
+    def add_template_values(self, mso_dict):
+        if self.template_id:
+            mso_dict["templateId"] = self.template_id
+        if self.template_name:
+            mso_dict["templateName"] = self.template_name
+        return mso_dict
+
     def get_object_by_key_value_pairs(self, object_description, search_list, kv_list, fail_module=False):
         """
         Get the object from a list of mso object dictionaries by name.
--- a/plugins/module_utils/template.py
+++ b/plugins/module_utils/template.py
@@ -95,13 +95,6 @@ class MSOTemplate:
         if self.template.get("templateType") != template_type:
             self.mso.fail_json(msg="Template type must be '{0}'.".format(template_type))
 
-    def add_template_values(self, mso_dict):
-        if self.template_id:
-            mso_dict["templateId"] = self.template_id
-        if self.template_name:
-            mso_dict["templateName"] = self.template_name
-        return mso_dict
-
     def get_object_by_key_value_pairs(self, object_description, search_list, kv_list, fail_module=False):
         """
         Get the object from a list of mso object dictionaries by name.
--- a/plugins/modules/ndo_node_profile.py
+++ b/plugins/modules/ndo_node_profile.py
@@ -55,7 +55,7 @@ options:
   nodes:
     description:
     - The list of node IDs to associate with the Node Profile.
-    - Each element can either be a single node ID or a range of IDs in quotation marks ('').
+    - Each element can either be a single node ID or a range of IDs.
     - This parameter is required when O(state=present).
     type: list
     elements: str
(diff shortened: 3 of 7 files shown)


plugins/modules/ndo_node_profile.py
```diff
@@ -0,0 +1,300 @@
...
+    description:
+    - The UUID of the Node Profile.
+    - This parameter is required when the O(name) needs to be updated.
+    type: str
+    aliases: [ node_profile_uuid ]
+  description:
+    description:
+    - The description of the Node Profile.
+    type: str
+  nodes:
```
akinross, line 49: should we support summarised input like 101-110, this is supported in UI, also for interfaces we supported something similar in ndo_port_channel_interface

plugins/modules/ndo_node_profile.py
```diff
@@ -0,0 +1,300 @@
...
+    description:
+    - The list of node IDs to associate with the Node Profile.
+    - This parameter is required when O(state=present).
```
akinross, line 52: should this then also include the or here?
```suggestion
    - This parameter or O(node_setting_uuid) is required when O(state=present).
```

akinross (changes requested):

plugins/modules/ndo_node_profile.py
samiib: `nodes` is always required when state is present. Only `node_setting` and `node_setting_uuid` are silently mutually exclusive as the UUID parameter takes precedence.

plugins/modules/ndo_node_profile.py
samiib: The UI appears to support a comma separated list of node ids (101,102,103), a range of ids (101-103) and both (101,102,103-105). It should be supported in the module already but I can add it as an example and test case.
`ndo_physical_interface` has a similar nodes parameter that supports the same cases.

plugins/modules/ndo_node_profile.py
akinross: you are right, I was looking at the wrong location thought this was the node_settings.

plugins/modules/ndo_node_profile.py
akinross: Yes think adding example and test case for this would be good idea

akinross (approved):
LGTM

akinross (dismissed):
LGTM

plugins/modules/ndo_node_profile.py
```diff
@@ -0,0 +1,301 @@
...
+    type: str
+    aliases: [ node_profile_uuid ]
+  description:
+    description:
+    - The description of the Node Profile.
+    type: str
+  nodes:
+    description:
+    - The list of node IDs to associate with the Node Profile.
+    - Each element can either be a single node ID or a range of IDs.
```
shrsr, line 52: - Each element can either be a single node ID or a range of IDs in quotation marks ('')

plugins/modules/ndo_node_profile.py
```diff
@@ -0,0 +1,301 @@
...
+    - This parameter is required when O(state=present).
+    type: list
+    elements: str
+  node_setting_uuid:
+    description:
+    - The UUID of the Node Setting to associate with the Node Profile.
```
shrsr, line 58: - The UUID of the node setting to associate with the Node Profile.

shrsr (changes requested):

plugins/modules/ndo_node_profile.py
```diff
@@ -0,0 +1,300 @@
...
+        argument_spec=argument_spec,
+        supports_check_mode=True,
+        required_if=[
+            ["state", "absent", ["name", "uuid"], True],
+            ["state", "present", ["name", "uuid"], True],
+            ["state", "present", ["node_setting_uuid", "node_setting"], True],
+            ["state", "present", ["nodes"]],
+        ],
+    )
+
```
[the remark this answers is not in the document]
samiib: I think we decided on not using mutually exclusive for uuid/name parameters. The UUID paramater will take precedence over the name/template dictionary instead.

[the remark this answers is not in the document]
samiib: Changed, as decided in stand up. 👍

plugins/modules/ndo_node_profile.py
```diff
@@ -0,0 +1,336 @@
...
+            mso.existing = mso_template.add_template_values(set_node_setting_relation(mso, match.details))  # When the state is present
+        else:
+            mso.existing = {}  # When the state is absent
+    elif module.check_mode and state != "query":  # When the state is present/absent with check mode
+        mso.existing = mso_template.add_template_values(set_node_setting_relation(mso, mso.proposed)) if state == "present" else {}
+
+    mso.exit_json()
+
+
+def set_node_setting_relation(mso, mso_dict):
```
shrsr, line 325: I created a general function in my PR for this based on a comments made related to it.
I'll need to modify it if we decide to use include template and template_id for references too..
You may need to use that function once I complete it.

shrsr (changes requested):

plugins/module_utils/mso.py
```diff
@@ -1553,6 +1553,21 @@ def input_validation(self, attr_name, attr_value, required_attributes, target_ob
         elif empty_attributes:
             self.module.fail_json(msg="When the '{0}' is '{1}', the {2} attributes must be set".format(attr_name, attr_value, empty_attributes))
 
+    def get_template_object_by_uuid(self, object_type, uuid, fail_module=True):
```
shrsr, line 1556: I would move this to utils file

plugins/module_utils/template.py
```diff
@@ -34,6 +34,8 @@ def __init__(self, mso_module, template_type=None, template_name=None, template_
             if self.template_summary:
                 self.template_path = "{0}/{1}".format(self.templates_path, self.template_id)
                 self.template = self.mso.query_obj(self.template_path)
+                self.template_name = self.template.get("templateName")
```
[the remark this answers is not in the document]
samiib: True! Thank you. Looks like the `templates/summaries` API has templateName but the `templates` API only has `displayName`.

plugins/module_utils/mso.py
samiib: Moved this to the MSOModule class for a few reasons.
1. It already requires us to pass in the MSOModule object if we add it to `utils.py`. This reduces the number of required arguments in the function.
2. The function is already available via the mso object so there is no need to import it from utils.
Please let me know if there is a reason to keep it in utils.py instead.

plugins/modules/ndo_node_profile.py
samiib: Thank you. I refactored my code to use the general function instead.

shrsr (approved):
LGTM

akinross (dismissed):
LGTM

plugins/modules/ndo_node_profile.py
```diff
@@ -0,0 +1,338 @@
...
+    type: str
+    aliases: [ node_profile_uuid ]
+  description:
+    description:
+    - The description of the Node Profile.
+    type: str
+  nodes:
+    description:
+    - The list of node IDs to associate with the Node Profile.
+    - Each element can either be a single node ID or a range of IDs in quotation marks ('').
```
shrsr, line 58: Sorry, looks like ('') is not required here. So reverting my previous comment.

- Each element can either be a single node ID or a range of IDs.

plugins/modules/ndo_node_profile.py
```diff
@@ -0,0 +1,338 @@
...
+  register: create_node_profile_1
+
+- name: Update the name and Node Setting of a Node Policy using UUID
+  cisco.mso.ndo_node_profile:
+    host: mso_host
+    username: admin
+    password: SomeSecretPassword
+    template_id: "{{ create_fabric_resource.current.templateId }}"
+    name: node_profile_1_updated
+    nodes: [101, 102, '103-105']
```
shrsr, line 128: No change required here but this can be:
nodes: [101, 102, 103-105]

plugins/modules/ndo_node_profile.py
```diff
@@ -0,0 +1,338 @@
...
+
+    mso = MSOModule(module)
+
+    template_name = module.params.get("template")
+    template_id = module.params.get("template_id")
+    name = module.params.get("name")
+    uuid = module.params.get("uuid")
+    description = module.params.get("description")
+    nodes = module.params.get("nodes")
+    if nodes:
```
shrsr, line 250: This snippet is not required:
```
if nodes:
        nodes = [str(node) for node in nodes]
```

shrsr (changes requested):

shrsr (dismissed):
LGTM

akinross (approved):
LGTM

gmicol (dismissed):
LGTM

plugins/modules/ndo_node_profile.py
```diff
@@ -0,0 +1,336 @@
...
+
+DOCUMENTATION = r"""
+---
+module: ndo_node_profile
+short_description: Manage Node Profiles on Cisco Nexus Dashboard Orchestrator (NDO).
+description:
+- Manage Node Profiles on Cisco Nexus Dashboard Orchestrator (NDO).
+- This module is only supported on ND v3.1 (NDO v4.3) and later.
+author:
+- Samita Bhattacharjee (@samiib)
```
anvitha-jain, line 26: Is it samiib or samitab ?

anvitha-jain (changes requested):

plugins/modules/ndo_node_profile.py
samiib: Yeah my GitHub username is @samiib.

akinross (approved):
LGTM

shrsr (approved):
LGTM

gmicol (approved):
LGTM

sajagana (approved):
LGTM!

anvitha-jain (approved):
LGTM

lhercot (approved):
LGTM

PR #615, opened 2025-02-24
Addition of new module ndo_l3out_interface_group_policy and its corresponding test file (DCNE-183)

shrsr:


plugins/modules/ndo_l3out_interface_group_policy.py
```diff
@@ -0,0 +1,512 @@
...
+        description:
+        - The BFD multi-hop key ID.
+        type: int
+      key:
+        description:
+        - The BFD multi-hop key.
+        type: str
+  ospf:
+    description:
+    - The OSPF configuration.
```
anvitha-jain, line 111: should we elaborate OSPF just at line 111?

plugins/modules/ndo_l3out_interface_group_policy.py
```diff
@@ -0,0 +1,512 @@
...
+    - Providing an empty dictionary will remove the O(interface_routing_policy={}) from L3Out Interface Group Policy.
+    type: dict
+    suboptions:
+      name:
+        description:
+        - The name of the L3Out Interface Routing Policy.
+        type: str
+      template:
+        description:
+        - The template associated with the  L3Out Interface Routing Policy.
```
anvitha-jain, line 58: remove extra space between --> the  L3Out

plugins/modules/ndo_l3out_interface_group_policy.py
```diff
@@ -0,0 +1,512 @@
...
+- name: Delete an existing L3Out interface group policy with name
+  cisco.mso.ndo_l3out_interface_group_policy:
+    host: mso_host
+    username: admin
+    password: SomeSecretPassword
+    template: l3out_template
+    l3out: l3out
+    name: interface_group_policy_1
+    state: absent
+"""
```
anvitha-jain, line 245: Examples for update, query and delete with UUID

plugins/modules/ndo_l3out_interface_group_policy.py
```diff
@@ -0,0 +1,512 @@
...
+        ospf=dict(
+            type="dict",
+            options=dict(
+                enable=dict(type="bool"),
+                authentication_type=dict(type="str", choices=["simple", "md5", "none"]),
+                key_id=dict(type="int"),
+                key=dict(type="str", no_log=True),
+            ),
+        ),
+        qos_priority=dict(type="str", choices=["level1", "level2", "level3", "level4", "level5", "level6", "unspecified"]),
```
anvitha-jain, line 308: if the values of QOS_LEVEL in constants.py is same as qos_priority, can we re-use this?

anvitha-jain (changes requested):

plugins/modules/ndo_l3out_interface_group_policy.py
shrsr: I think we have sub options here which explain more..

plugins/modules/ndo_l3out_interface_group_policy.py
shrsr: There's no UUID associated with this object

plugins/modules/ndo_l3out_interface_group_policy.py
```diff
@@ -0,0 +1,538 @@
...
+  description:
+    description:
+    - The description of the L3Out Interface Group Policy.
+    type: str
+  name:
+    description:
+    - The name of the L3Out Interface Group Policy.
+    type: str
+    aliases: [ l3out_interface_group_policy ]
+  interface_routing_policy:
```
sajagana, line 53: The `Interface Routing Policy` is unique within the tenant level, so only `Interface Routing Policy` name is enough. Please recheck the logic once again.

plugins/modules/ndo_l3out_interface_group_policy.py
```diff
@@ -0,0 +1,538 @@
...
+      key:
+        description:
+        - The OSPF key.
+        type: str
+  qos_priority:
+    description:
+    - The QoS priority level.
+    type: str
+    choices: [ level1, level2, level3, level4, level5, level6, unspecified ]
+  custom_qos_policy:
```
sajagana, line 144: The `Custom QoS Policy` is unique within the tenant level, so only `Custom QoS Policy` name is enough. Please recheck the logic once again.

plugins/modules/ndo_l3out_interface_group_policy.py
```diff
@@ -0,0 +1,538 @@
...
+        )
+        if match:
+            for attribute_name in attributes_dict.keys():
+                if match.details.get(attribute_name, {}).get("key", {}).get("ref"):
+                    match.details[attribute_name]["key"].pop("ref", None)
+            mso.existing = match.details
+        else:
+            mso.existing = {}  # When the state is absent
+    elif module.check_mode and state != "query":  # When the state is present/absent with check mode
+        mso.existing = mso.proposed if state == "present" else {}
```
sajagana, line 491: Insert the `Interface Routing Policy` and `Custom QoS Policy` object names in the `current` and `previous` portion of the module output when we do `create, update, query and delete`.

plugins/modules/ndo_l3out_interface_group_policy.py
```diff
@@ -0,0 +1,538 @@
...
+    if name:
+        match = mso_template.get_object_by_key_value_pairs(
+            object_description,
+            existing_interface_groups,
+            [KVPair("name", name)],
+        )
+        if match:
+            mso.existing = mso.previous = copy.deepcopy(match.details)
+    else:
+        mso.existing = mso.previous = existing_interface_groups
```
sajagana, line 389: Insert the `Interface Routing Policy` and `Custom QoS Policy` object names in the `current` and `previous` portion of the module output when we do `create, update, query and delete`.

plugins/module_utils/template.py
```diff
@@ -291,3 +291,31 @@ def get_l3out_interface_routing_policy_object(self, uuid=None, name=None, fail_m
                 fail_module,
             )
         return existing_l3out_interface_routing_policy  # Query all objects
+
+    def get_l3out_interface_routing_policy_uuid(self, mso_module, interface_routing_policy_name, template):
+        """
+        Get the UUID of an L3Out Interface Routing Policy by name.
+        :param interface_routing_policy_name: Name of the L3Out Interface Routing Policy to search for -> Str
+        :return: UUID of the Interface Routing Policy -> Str
+        """
```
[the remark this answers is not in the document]
shrsr: Can you check if what I have now is what you asked for?

plugins/modules/ndo_l3out_interface_group_policy.py
shrsr: OK

plugins/modules/ndo_l3out_interface_group_policy.py
```diff
@@ -0,0 +1,562 @@
...
+                if attribute_dict is not None:
+                    attribute_is_empty = check_if_all_elements_are_none(list(attribute_dict.values()))
+
+                    if match.details.get(attribute_name, {}).get("key", {}).get("ref"):
+                        match.details[attribute_name]["key"].pop("ref", None)
+                        mso.previous[attribute_name]["key"].pop("ref", None)
+                        mso.existing[attribute_name]["key"].pop("ref", None)
+
+                    if attribute_is_empty and match.details.get(attribute_name):
+                        mso_values_remove.append(attribute_name)
```
akinross, line 444: just so I understand, this is required for the removal of authentication for a specific protocol?

plugins/modules/ndo_l3out_interface_group_policy.py
```diff
@@ -0,0 +1,562 @@
...
+                    elif not attribute_is_empty and not match.details.get(attribute_name):
+                        update_mso_values_attributes_payload(mso_values, {attribute_name: attribute_dict})
+
+                    elif not attribute_is_empty and match.details.get(attribute_name):
+                        mso_values[(attribute_name, "enabled")] = attribute_dict.get("enable")
+                        if attribute_dict.get("enable") is True:
+                            if attribute_name == "ospf":
+                                mso_values[(attribute_name, "authType")] = attribute_dict.get("authentication_type")
+                            else:
+                                mso_values[(attribute_name, "authEnabled")] = attribute_dict.get("authentication")
```
akinross, line 455: in sabari his previous node group he uses ENABLED_OR_DISABLED_TO_BOOL_STRING_MAP is that also needed here? might also apply to the other bool value

plugins/modules/ndo_l3out_interface_group_policy.py
```diff
@@ -0,0 +1,562 @@
...
+                        mso.existing[attribute_name]["key"].pop("ref", None)
+
+                    if attribute_is_empty and match.details.get(attribute_name):
+                        mso_values_remove.append(attribute_name)
+
+                    elif not attribute_is_empty and not match.details.get(attribute_name):
+                        update_mso_values_attributes_payload(mso_values, {attribute_name: attribute_dict})
+
+                    elif not attribute_is_empty and match.details.get(attribute_name):
+                        mso_values[(attribute_name, "enabled")] = attribute_dict.get("enable")
```
akinross, line 450: Why are we sometimes calling update_mso_values_attributes_payload ( also for non existing ) and we have other logic below. Is the creation of mso_values for these protocol with authentication settings not always equal?

With this comment I do not mean the removal of config.

plugins/modules/ndo_l3out_interface_group_policy.py
shrsr: This for removal of both the protocol and the authentication of a protocol. Without the protocol there's no authentication required for it. (example -> bfd: {})

plugins/modules/ndo_l3out_interface_group_policy.py
```diff
@@ -0,0 +1,562 @@
...
+            mso.existing = mso.previous = copy.deepcopy(match.details)
+    else:
+        mso.existing = mso.previous = [set_names_for_references(mso, interface_group) for interface_group in existing_interface_groups]
+
+    if state != "query":
+        interface_group_policy_path = "/l3outTemplate/l3outs/{0}/interfaceGroups/{1}".format(l3out_object.index, match.index if match else "-")
+
+    ops = []
+
+    attributes_dict = {"bfd": bfd, "bfdMultiHop": bfd_multi_hop, "ospf": ospf}
```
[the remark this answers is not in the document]
shrsr: I think protocol_dict would be better. Changed it!

plugins/modules/ndo_l3out_interface_group_policy.py
```diff
@@ -0,0 +1,562 @@
...
+                    mso_values[attribute_name].update(
+                        {
+                            "authType": attribute_dict.get("authentication_type"),
+                            "keyID": attribute_dict.get("key_id"),
+                            "key": {"value": attribute_dict.get("key")},
+                        }
+                    )
+
+
+def set_names_for_references(mso, interface_group_policy_dict):
```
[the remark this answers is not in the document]
shrsr: It can be. However, it is unlikely that we'll come up with another module modifying a dict with the exact same references.

plugins/modules/ndo_l3out_interface_group_policy.py
```diff
@@ -0,0 +1,562 @@
...
+    module = AnsibleModule(
+        argument_spec=argument_spec,
+        supports_check_mode=True,
+        required_if=[
+            ["state", "present", ["template", "template_id"], True],
+            ["state", "query", ["template", "template_id"], True],
+            ["state", "absent", ["template", "template_id"], True],
+            ["state", "present", ["l3out", "l3out_uuid"], True],
+            ["state", "query", ["l3out", "l3out_uuid"], True],
+            ["state", "absent", ["l3out", "l3out_uuid"], True],
```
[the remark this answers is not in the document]
shrsr: Yeah, that makes sense! I'll change them

plugins/modules/ndo_l3out_interface_group_policy.py
shrsr: It's not needed here because I directly use bool as input and pass it to the payload. I will change it if it becomes a consistency issue?

plugins/modules/ndo_l3out_interface_group_policy.py
```diff
@@ -0,0 +1,562 @@
...
+        mutually_exclusive=[
+            ["interface_routing_policy", "interface_routing_policy_uuid"],
+            ["custom_qos_policy", "custom_qos_policy_uuid"],
+        ],
+    )
+
+    mso = MSOModule(module)
+
+    template_identifier = get_object_identifier(module.params.get("template_id"), module.params.get("template"))
+    l3out_identifier = get_object_identifier(module.params.get("l3out_uuid"), module.params.get("l3out"))
```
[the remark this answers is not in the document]
shrsr: I agree. I picked it up from another PR but made the changes.

plugins/modules/ndo_l3out_interface_group_policy.py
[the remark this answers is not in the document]
gmicol: I think it should be a generic function, so far we've seen that the type of object which are supported for this query endpoint all have name and ref attached (just the name is not returned by the API) so we could potentially do something like this:
```PYTHON
map = {
    "qos": {"name": "qosName", "ref": "qosRef"},
    "l3OutIntfPolGroup": {"name": "interfaceRoutingPolicyName", "ref": "interfaceRoutingPolicyRef"},
}

def set_names_for_references(mso, interface_group_policy_dict, map):
    for object_type, values in map.items():
      if interface_group_policy_dict.get(values.get("ref")):
          interface_group_policy_dict[values.get("name")] = get_template_object_name_by_uuid(mso, object_type, interface_group_policy_dict.get(values.get("ref")))
    return interface_group_policy_dict
```
it's definitely not optimal (the structure of the map could be completely different for example and the name of the variable should be changed, etc...) but the idea is there, what do you think?

gmicol (changes requested):

plugins/modules/ndo_l3out_interface_group_policy.py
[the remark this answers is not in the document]
shrsr: We could come up with a general function such as above and apply it to all the modules..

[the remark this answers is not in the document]
shrsr: Let's post this in standup space to make sure we all use it once is is approved

plugins/modules/ndo_l3out_interface_group_policy.py
```diff
@@ -0,0 +1,563 @@
...
+    - Determines the desired state of the resource.
+    - Use C(absent) to remove the resource.
+    - Use C(query) to list the resource.
+    - Use C(present) to create or update the resource.
+    type: str
+    choices: [ absent, query, present ]
+    default: query
+notes:
+- The O(template) must exist before using this module in your playbook.
+  Use M(cisco.mso.ndo_template) to create the L3Out template.
```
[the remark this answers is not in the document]
shrsr: The referenced objects are not required. Hence, the wording is a little different if you notice..

gmicol (approved):
LGTM

plugins/module_utils/template.py
```diff
@@ -291,3 +291,35 @@ def get_l3out_interface_routing_policy_object(self, uuid=None, name=None, fail_m
...
+            "L3Out Interface Routing Policyt", existing_l3out_interface_routing_policies, [KVPair("name", interface_routing_policy_name)], fail_module=True
+        )
+        return match.details.get("uuid")
+
+    def get_custom_qos_policy_uuid(self, mso_module, custom_qos_policy_name, template):
+        """
+        Get the UUID of a Custom QoS Policy by name.
+        :param mso_module: An instance of mso module calling this function -> MSOModule
+        :param custom_qos_policy_name: Name of the Custom QoS Policy to search for -> Str
+        :param template: Template asscociated with custom_qos_policy_name -> Str
```
anvitha-jain, line 316: asscociated -> associated

plugins/module_utils/template.py
```diff
@@ -291,3 +291,35 @@ def get_l3out_interface_routing_policy_object(self, uuid=None, name=None, fail_m
...
+        :param mso_module: An instance of mso module calling this function -> MSOModule
+        :param interface_routing_policy_name: Name of the L3Out Interface Routing Policy to search for -> Str
+        :param template: Template asscociated with interface_routing_policy_name -> Str
+        :return: UUID of the Interface Routing Policy -> Str
+        """
+        tenant_template = MSOTemplate(mso_module, "tenant", template)
+        tenant_template.validate_template("tenantPolicy")
+        existing_l3out_interface_routing_policies = tenant_template.template.get("tenantPolicyTemplate", {}).get("template", {}).get("l3OutIntfPolGroups", [])
+        match = self.get_object_by_key_value_pairs(
+            "L3Out Interface Routing Policyt", existing_l3out_interface_routing_policies, [KVPair("name", interface_routing_policy_name)], fail_module=True
```
anvitha-jain, line 307: Policyt -> Policy

plugins/module_utils/template.py
```diff
@@ -291,3 +291,35 @@ def get_l3out_interface_routing_policy_object(self, uuid=None, name=None, fail_m
                 fail_module,
             )
         return existing_l3out_interface_routing_policy  # Query all objects
+
+    def get_l3out_interface_routing_policy_uuid(self, mso_module, interface_routing_policy_name, template):
+        """
+        Get the UUID of an L3Out Interface Routing Policy by name.
+        :param mso_module: An instance of mso module calling this function -> MSOModule
+        :param interface_routing_policy_name: Name of the L3Out Interface Routing Policy to search for -> Str
+        :param template: Template asscociated with interface_routing_policy_name -> Str
```
anvitha-jain, line 300: asscociated -> associated

plugins/module_utils/utils.py
```diff
@@ -143,6 +143,35 @@ def check_if_all_elements_are_none(values):
...
+    if response_object:
+        return response_object.get("name")
+
+
+def set_names_for_references(mso, payload, reference_dict):
+    """
+    Return the updated payload with names for references.
+    :param mso: An instance of the MSO class, which provides methods for making API requests -> MSO Class instance
+    :param payload: The original payload that requires to be updated  -> Dict
+    :param eference_dict: A dict containing references and the corresponding names -> Dict
```
anvitha-jain, line 171: eference_dict -> reference_dict

anvitha-jain (dismissed):
LGTM

gmicol (approved):
LGTM

anvitha-jain (approved):
LGTM

anvitha-jain (dismissed):
LGTM

samiib (changes requested):

plugins/module_utils/utils.py
```diff
@@ -143,6 +143,61 @@ def check_if_all_elements_are_none(values):
...
+    response_object = mso.request("templates/objects?type={0}&uuid={1}".format(object_type, uuid), "GET")
+    if not response_object and fail_module:
+        msg = "Provided {0} with UUID of '{1}' not found.".format(object_type, uuid)
+        mso.fail_json(msg=msg)
+    return response_object
+
+
+def set_names_for_references(mso, payload, reference_dict):
+    """
+    Return the updated payload with names for references.
```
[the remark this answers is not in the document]
shrsr: Sure!

plugins/module_utils/utils.py
```diff
@@ -143,6 +143,61 @@ def check_if_all_elements_are_none(values):
     Checks if all the elements in the provided list are None.
 
     :param values: List of values to check. -> List
-    :return: True if all elements are None, False otherwise. -> boo
+    :return: True if all elements are None, False otherwise. -> bool
     """
     return all(value is None for value in values)
+
+
+def get_template_object_name_by_uuid(mso, object_type, uuid):
```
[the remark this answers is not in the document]
shrsr: We could move them but our mso.py is humongous.
My opinion:
https://github.com/CiscoDevNet/ansible-mso/pull/614/files#r1983346217

gmicol (dismissed):
LGTM

plugins/module_utils/utils.py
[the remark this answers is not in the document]
gmicol: I think we should either keep it in utils.py because as Shreyas said mso.py is humongous and our auxiliary functions should be separated from the main code of mso which makes it cleaner and more readable.

But we could also move this function in templates.py as it would make more sense to have this function there, what do you think @shrsr ?

I am actually good with both options.

anvitha-jain (dismissed):
LGTM

gmicol (approved):
LGTM

plugins/modules/ndo_l3out_interface_group_policy.py
```diff
@@ -0,0 +1,630 @@
...
+        },
+    }
+    if name:
+        match = mso_template.get_object_by_key_value_pairs(
+            object_description,
+            existing_interface_groups,
+            [KVPair("name", name)],
+        )
+        if match:
+            mso_template.set_template_id_and_name(mso_template.set_names_for_references(mso, match.details, reference_dict))
```
samiib, line 430: I wonder if we could merge these two functions into another utility function that takes in an optional `reference_dict` to simplify our module code?

tests/integration/targets/ndo_l3out_interface_group_policy/tasks/main.yml
```diff
@@ -0,0 +1,604 @@
...
+
+    - name: Create L3Out interface group policy object with default values - normal mode
+      cisco.mso.ndo_l3out_interface_group_policy: &nm_interface_group_policy_1_present
+        <<: *cm_interface_group_policy_1_present
+      register: nm_interface_group_policy_1_present
+
+    - name: Assertion check for create L3Out interface group policy object with default values - normal mode
+      ansible.builtin.assert:
+        that:
+          - nm_interface_group_policy_1_present is changed
```
samiib, line 183: Should we add an assertion for `templateName` and `templateId` in all the test cases?

samiib (changes requested):

plugins/module_utils/template.py
```diff
@@ -291,3 +300,140 @@ def get_l3out_interface_routing_policy_object(self, uuid=None, name=None, fail_m
...
+        :param tenant_template: The tenant object -> Dict
+        :param policy_name: Name of the policy -> Str
+        :param policy_type: The type of the policy specified in the API response -> Str
+        :return: UUID of the tenant policy -> Str
+        """
+        existing_policies = tenant_template.template.get("tenantPolicyTemplate", {}).get("template", {}).get(policy_type, [])
+        match = self.get_object_by_key_value_pairs(policy_type, existing_policies, [KVPair("name", policy_name)], fail_module=True)
+        return match.details.get("uuid")
+
+    @staticmethod
```
sajagana, line 316: Why is it set to `staticmethod`? Same question for other `staticmethod`.

Also we can access the `mso` directly from the `MSOTemplate` using `self.mso`.

sajagana (changes requested):

tests/integration/targets/ndo_l3out_interface_group_policy/tasks/main.yml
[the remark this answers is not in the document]
shrsr: This comment is not clear?

plugins/modules/ndo_l3out_interface_group_policy.py
```diff
@@ -0,0 +1,630 @@
...
+  Use M(cisco.mso.ndo_tenant_custom_qos_policy) to create the Custom QoS Policy.
+seealso:
+- module: cisco.mso.ndo_template
+- module: cisco.mso.ndo_l3out_template
+- module: cisco.mso.ndo_l3out_interface_routing_policy
+- module: cisco.mso.ndo_tenant_custom_qos_policy
+extends_documentation_fragment: cisco.mso.modules
+"""
+
+EXAMPLES = r"""
```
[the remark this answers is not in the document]
shrsr: This module has no UUID..

anvitha-jain (dismissed):
LGTM

plugins/modules/ndo_l3out_interface_group_policy.py
```diff
@@ -0,0 +1,629 @@
...
+                template=dict(type="str"),
+                template_id=dict(type="str"),
+            ),
+            required_by={
+                "template": "name",
+                "template_id": "name",
+            },
+            mutually_exclusive=[
+                ["template", "template_id"],
+            ],
```
samiib, line 312: This sub reference spec section can become a common dict in either mso.py or template.py.
We may all use the same spec for the name/template reference object parameter.
Also it can change to this instead:
```suggestion
        interface_routing_policy=dict(
            type="dict",
            options=dict(
                name=dict(type="str", required=True),
                template=dict(type="str"),
                template_id=dict(type="str"),
            ),
            required_one_of=[
                ["template", "template_id"],
            ],
            mutually_exclusive=[
                ("template", "template_id"),
            ],
```
Then we wouldn't need the `check_template_when_name_is_provided` function as the spec does the check for us.

samiib (changes requested):

samiib (dismissed):
LGTM

sajagana (approved):
LGTM!

plugins/modules/ndo_l3out_interface_group_policy.py
```diff
@@ -0,0 +1,629 @@
...
+            mso.existing = {}  # When the state is absent
+    elif module.check_mode and state != "query":  # When the state is present/absent with check mode
+        mso_template.update_config_with_template_and_references(mso.proposed, reference_dict)
+        mso.existing = mso.proposed if state == "present" else {}
+
+    mso.exit_json()
+
+
+def check_authentication_requirements(mso, protocols_dict):
+    failed_attributes = {}
```
akinross, line 578: not sure I understand the dict choice here? why not a list or directly a string, when keys are not used?

plugins/modules/ndo_l3out_interface_group_policy.py
```diff
@@ -0,0 +1,629 @@
...
+                    mso_values[protocol_name].update(
+                        {
+                            "authType": protocol_dict.get("authentication_type"),
+                            "keyID": protocol_dict.get("key_id"),
+                            "key": {"value": protocol_dict.get("key")},
+                        }
+                    )
+
+
+def process_policy(mso_values, mso_template, tenant_template, policy, policy_ref, policy_type):
```
akinross, line 624: is this a generic function that could be reused in other modules? 

should we rename it `process_tenant_policy` since it is specific for tenant templates?

akinross (changes requested):

akinross (approved):
LGTM

gmicol (approved):
LGTM

anvitha-jain (approved):
LGTM

gmicol (dismissed):
LGTM

samiib (dismissed):
LGTM

akinross (dismissed):
LGTM

sajagana (approved):
LGTM!

samiib (approved):
LGTM

akinross (approved):
LGTM

gmicol (approved):
LGTM

sajagana (approved):
LGTM!

anvitha-jain (approved):
LGTM

lhercot (approved):
LGTM

PR #551, opened 2024-10-11
Add a new module to manage Virtual Port Channel Interfaces in Fabric Resource Policies Template. (DCNE-53)

gmicol:
resolves #467 

New Modules Added:
- ndo_virtual_port_channel_interface

PR #623, opened 2025-03-03
Addition of a new module ndo_qos_dscp_cos_translation_policy and its test file (DCNE-72)

shrsr:


plugins/modules/ndo_qos_dscp_cos_translation_policy.py
```diff
@@ -0,0 +1,553 @@
...
+    description:
+    - Use C(absent) for removing.
+    - Use C(query) for listing an object or multiple objects.
+    - Use C(present) for creating or updating.
+    type: str
+    choices: [ absent, query, present ]
+    default: query
+notes:
+- The O(template) must exist before using this module in your playbook.
+  The M(cisco.mso.ndo_template) module can be used for this.
```
gmicol, line 347: Shouldn't we add a extra note like this:
https://github.com/CiscoDevNet/ansible-mso/pull/551/files#diff-94f14791e8b830ecce8e5f4b333e1bfa050b749b4b160ab77b6484e0046a9dd4R143

Or is not needed to specify this to the user?

gmicol (changes requested):

gmicol (approved):
LGTM

anvitha-jain (dismissed):
LGTM

plugins/modules/ndo_qos_dscp_cos_translation_policy.py
```diff
@@ -0,0 +1,555 @@
...
+    "status": ["preview"],
+    "supported_by": "community",
+}
+
+DOCUMENTATION = r"""
+---
+module: ndo_qos_dscp_cos_translation_policy
+short_description: Manage QoS DSCP CoS Translation Policies on Cisco Nexus Dashboard Orchestrator (NDO).
+description:
+- Manage Quality of Service (QoS) Differentiated Services Code Point (DSCP) Class of Service (CoS) Translation Policies.
```
samiib, line 23: Looks like users can only configure one QoS DSCP CoS Translation Policy per template. There is a similar restriction on MCP Global Policies. I added a few notes to indicate this and that updating a policy with a different name would only update the existing policy instead of adding a new one. eg [L20](https://github.com/CiscoDevNet/ansible-mso/blob/master/plugins/modules/ndo_mcp_global_policy.py#L20) and [L111](https://github.com/CiscoDevNet/ansible-mso/blob/master/plugins/modules/ndo_mcp_global_policy.py#L111)

Should we use similar notes and descriptions in this module?

samiib (changes requested):

plugins/modules/ndo_qos_dscp_cos_translation_policy.py
shrsr: Sounds good to me!

anvitha-jain (approved):
LGTM

lhercot (approved):
LGTM

PR #656, opened 2025-04-21
Added ndo_fabric_span_session module for the Fabric Monitoring Access Policy template (DCNE-138)

sajagana:

PR #639, opened 2025-04-08
[minor_change] Added new module for Tenant SPAN Session (DCNE-136)

samiib:
Fixes #462

PR #632, opened 2025-03-28
Addition of new module ndo_service_device_cluster and its test file (DCNE-137)

shrsr:

PR #672, opened 2025-05-16
[minor_change] Add module ndo_pod_profile for pod profiles in fabric resource templates (DCNE-141)

### LINKED ISSUE #469: Enhancement: New module for Pod Profile in Fabric Resource Policies Template (DCNE-141)
### Community Note

### Description

Create new module for Pod Profile in Fabric Resource Policies Template ( created via ndo_template module )

* mso_XXXXX

* ND 3.1.1 / NDO 4.3.1

### Collection versions

* cisco.mso x.x.x

### References

* #0000

akinross:
fixes #469

### DIFF
--- a/plugins/module_utils/template.py
+++ b/plugins/module_utils/template.py
@@ -79,6 +79,12 @@ class MSOTemplate:
                  Values of provided keys of all existing objects. -> List
         """
 
+        # Sometimes the attribute returned by api might be None
+        # If search_list is None, iterating over it will throw an error
+        # Thus we need to return the match of None and without existing values
+        if search_list is None:
+            return None, []
+
         def kv_match(kvs, item):
             return all((item.get(kv.key) == kv.value for kv in kvs))
 
@@ -293,6 +299,26 @@ class MSOTemplate:
         existing_objects = self.template.get("fabricPolicyTemplate", {}).get("template", {}).get("nodePolicyGroups", [])
         if uuid or name:  # Query a specific object
             return self.get_object_by_key_value_pairs("Node Settings", existing_objects, [KVPair("uuid", uuid) if uuid else KVPair("name", name)], fail_module)
+        return existing_objects  # Query all objects
+
+    def get_pod_profile_object(self, uuid=None, name=None, search_object=None, fail_module=False):
+        """
+        Get the Pod Profile by uuid or name.
+        :param uuid: UUID of the Pod Profile to search for -> Str
+        :param name: Name of the Pod Profile to search for -> Str
+        :param search_object: The object to search in -> Dict
+        :param fail_module: When match is not found fail the ansible module -> Bool
+        :return: Dict | None | List[Dict] | List[]: The processed result which could be:
+                 When the UUID | Name is existing in the search list -> Dict
+                 When the UUID | Name is not existing in the search list -> None
+                 When both UUID and Name are None, and the search list is not empty -> List[Dict]
+                 When both UUID and Name are None, and the search list is empty -> List[]
+        """
+        if not search_object:
+            search_object = self.template
+        existing_objects = search_object.get("fabricResourceTemplate", {}).get("template", {}).get("podProfiles", [])
+        if uuid or name:  # Query a specific object
+            return self.get_object_by_key_value_pairs("Pod Profile", existing_objects, [KVPair("uuid", uuid) if uuid else KVPair("name", name)], fail_module)
         return existing_objects  # Query all objects
 
     def get_pod_settings_object(self, uuid=None, name=None, search_object=None, fail_module=False):
--- a/plugins/modules/ndo_pod_profile.py
+++ b/plugins/modules/ndo_pod_profile.py
@@ -1,0 +1,353 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# Copyright: (c) 2025, Akini Ross (@akinross) <akinross@cisco.com>
+
+# GNU General Public License v3.0+ (see LICENSE or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+
+__metaclass__ = type
+
+ANSIBLE_METADATA = {"metadata_version": "1.1", "status": ["preview"], "supported_by": "community"}
+
+DOCUMENTATION = r"""
+---
+module: ndo_pod_profile
+short_description: Manage Pod Profiles on Cisco Nexus Dashboard Orchestrator (NDO).
+description:
+- Manage Pod Profiles on Cisco Nexus Dashboard Orchestrator (NDO).
+- This module is only supported on ND v3.2 (NDO v4.4) and later.
+author:
+- Akini Ross (@akinross)
+options:
+  template:
+    description:
+    - The name of the template.
+    - The template must be a Fabric Resource Policy template.
+    - This parameter or O(template_id) is required.
+    type: str
+  template_id:
+    description:
+    - The ID of the template.
+    - The template must be a Fabric Resource Policy template.
+    - This parameter or O(template) is required.
+    type: str
+  name:
+    description:
+    - The name of the Pod Profile.
+    type: str
+    aliases: [ pod_profile ]
+  uuid:
+    description:
+    - The UUID of the Pod Profile.
+    - This parameter is required when the O(name) needs to be updated.
+    type: str
+    aliases: [ pod_profile_uuid ]
+  description:
+    description:
+    - The description of the Pod Profile.
+    type: str
+  pods:
+    description:
+    - A list of pod IDs to be used in the Pod Profile.
+    - When provided, the Pod Profile will apply to pods specified in the range.
+    - When an empty list O(pods=[]) is provided, the Pod Profile will apply to all pods.
+    - When not provided on inital create, the Pod Profile will apply to all pods.
+    type: list
+    elements: int
+    aliases: [ blocks ]
+  pod_settings:
+    description:
+    - The name of the Pod Settings Policy to be used.
+    - Providing an empty string O(pod_settings="") will remove the Pod Settings Policy from the Pod Profile.
+    type: str
+    aliases: [ pod_settings_name ]
+  pod_settings_uuid:
+    description:
+    - The UUID of the Pod Settings Policy to be used.
+    - Providing an empty string O(pod_settings_uuid="") will remove the Pod Settings Policy from the Pod Profile.
+    type: str
+  pod_settings_template:
+    description:
+    - The name of the Pod Settings template.
+    - This parameter or O(pod_settings_template_id) is required when O(pod_settings) or O(pod_settings_uuid) is used.
+    type: str
+  pod_settings_template_id:
+    description:
+    - The ID of the Pod Settings template.
+    - This parameter or O(pod_settings_template) is required when O(pod_settings) or O(pod_settings_uuid) is used.
+    type: str
+  state:
+    description:
+    - Use C(absent) for removing.
+    - Use C(query) for listing an object or multiple objects.
+    - Use C(present) for creating or updating.
+    type: str
+    choices: [ absent, query, present ]
+    default: query
+notes:
+- The O(template) must exist before using this module in your playbook.
+  Use M(cisco.mso.ndo_template) to create the Fabric Resource Policy template.
+- The O(pod_settings) or O(pod_settings_uuid) must exist before using it with this module in your playbook.
+  Use M(cisco.mso.ndo_pod_settings) to create the Pod Settings Policy.
+seealso:
+- module: cisco.mso.ndo_template
+- module: cisco.mso.ndo_pod_settings
+extends_documentation_fragment: cisco.mso.modules
+"""
+
+EXAMPLES = r"""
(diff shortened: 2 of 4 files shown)


sajagana (approved):
LGTM!

plugins/modules/ndo_pod_profile.py
```diff
@@ -0,0 +1,353 @@
...
+- name: Update a Pod Profile using UUID
+  cisco.mso.ndo_pod_profile:
+    host: mso_host
+    username: admin
+    password: SomeSecretPassword
+    template: ansible_test
+    name: ansible_pod_profile_changed
+    uuid: "{{ create_pod_profile.current.uuid }}"
+    pod_settings_uuid: "{{ create_pod_settings.current.uuid }}"
+    state: present
```
gmicol, line 135: Missing `pod_settings_template` (or `pod_settings_template_id`) as it is required with `pod_settings` (or `pod_settings_uuid`)

gmicol (changes requested):

plugins/modules/ndo_pod_profile.py
```diff
@@ -0,0 +1,356 @@
...
+            mso.existing = match.details  # When the state is present
+        else:
+            mso.existing = {}  # When the state is absent
+    elif module.check_mode and state != "query":  # When the state is present/absent with check mode
+        set_pod_profile_policy_names(mso, mso_templates, mso.proposed)
+        mso.existing = mso.proposed if state == "present" else {}
+    mso.exit_json()
+
+
+def set_pod_profile_policy_names(mso, mso_templates, pod_profile):
```
shrsr, line 314: This function is really useful for policy objects that are not supported by the API yet. Can we make it generic?
It will be useful for future modules to consume this function.

plugins/module_utils/template.py
```diff
@@ -79,6 +79,12 @@ def get_object_from_list(search_list, kv_list):
                  Values of provided keys of all existing objects. -> List
         """
 
+        # Sometimes the attribute returned by api might be None
+        # If search_list is None, iterating over it will throw an error
+        # Thus we need to return the match of None and without existing values
+        if search_list is None:
```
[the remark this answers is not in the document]
akinross: This should not impact the existing module using it, it would only prevent an traceback error when the searchlist is None. 

NDO API is inconsistent in return of values, some are empty list, some are not provided ( resulting in our code to empty list, and some are null. The null case would result in the the error, thus to avoid this we should return empty list. We could do this at the function level but think it is easier to do the check once here.

plugins/modules/ndo_pod_profile.py
```diff
@@ -0,0 +1,356 @@
...
+
+        mso_values = {
+            "name": name,
+            "description": description,
+        }
+
+        if pods:
+            mso_values["blocks"] = pods
+            mso_values["kind"] = "podRange"
+        elif pods == []:
```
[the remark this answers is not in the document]
akinross: If we do not specify we stick to the default behaviour of NDO, thus only when you set [] we set to all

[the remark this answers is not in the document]
akinross: this is for the update case to be set correct where we do not want to change when not provided, only on the create the we force it to all when not provided

plugins/modules/ndo_pod_profile.py
```diff
@@ -0,0 +1,356 @@
...
+        if match:
+            remove_data = []
+            if pods == []:
+                remove_data.append("blocks")
+
+            append_update_ops_data(ops, match.details, pod_profile_path, mso_values, remove_data)
+            mso.sanitize(mso_values, collate=True, unwanted=remove_data)
+            set_pod_profile_policy_names(mso, mso_templates, mso.proposed)
+        else:
+            if not pods:
```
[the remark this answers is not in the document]
akinross: see other comment

plugins/modules/ndo_pod_profile.py
akinross: For now I chose to include in the module since I did not want to change existing function. Shall we create a separate issue to track this feature?

plugins/modules/ndo_pod_profile.py
```diff
@@ -0,0 +1,353 @@
...
+        }
+
+        if pods:
+            mso_values["blocks"] = pods
+            mso_values["kind"] = "podRange"
+        elif pods == []:
+            mso_values["kind"] = "all"
+
+        pod_settings_mso_template = mso_templates.get_template("fabric_policy", pod_settings_template, pod_settings_template_id)
+        mso_values["policy"] = pod_settings_mso_template.get_pod_settings_object(pod_settings_uuid, pod_settings, fail_module=True).details.get("uuid")
```
sajagana, line 280: Can we ignore the pod settings object lookup when the "pod_settings_uuid" is provided by the user?

sajagana (changes requested):

plugins/modules/ndo_pod_profile.py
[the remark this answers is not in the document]
shrsr: I know this is for the update case. I was thinking should we make the behaviour consistent for both create and update where if pods is not provided means the profile is applied to all. 
Let's skip this if unnecessary..

plugins/modules/ndo_pod_profile.py
shrsr: Yes, let's open another issue then

plugins/module_utils/template.py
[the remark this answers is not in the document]
shrsr: I think we already handle this in the modules by using : return_value or []

plugins/modules/ndo_pod_profile.py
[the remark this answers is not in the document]
akinross: ok i thought we would never update anything that was not provided?

plugins/module_utils/template.py
```diff
@@ -295,6 +301,86 @@ def get_node_settings_object(self, uuid=None, name=None, fail_module=False):
...
+                 When the UUID | Name is existing in the search list -> Dict
+                 When the UUID | Name is not existing in the search list -> None
+                 When both UUID and Name are None, and the search list is not empty -> List[Dict]
+                 When both UUID and Name are None, and the search list is empty -> List[]
+        """
+        if not search_object:
+            search_object = self.template
+        existing_objects = search_object.get("fabricPolicyTemplate", {}).get("template", {}).get("podPolicyGroups", [])
+        if uuid or name:  # Query a specific object
+            return self.get_object_by_key_value_pairs("Pod Settings", existing_objects, [KVPair("uuid", uuid) if uuid else KVPair("name", name)], fail_module)
```
samiib, line 341: Should this be the plural `Pod Settings` or singular `Pod Setting`.

plugins/modules/ndo_pod_profile.py
```diff
@@ -0,0 +1,357 @@
...
+  pods:
+    description:
+    - A list of pod IDs to be used in the Pod Profile.
+    - When provided, the Pod Profile will apply to pods specified in the range.
+    - When an empty list O(pods=[]) is provided, the Pod Profile will apply to all pods.
+    - When not provided on inital create, the Pod Profile will apply to all pods.
+    type: list
+    elements: int
+    aliases: [ blocks ]
+  pod_settings:
```
samiib, line 60: We have previously been nesting this reference objects into dictionaries. 
ie.
```
pod_settings: dictionary {
   name: str, required,
   template: str,
   template_id: str
}
pod_settings_uuid: str
```
Have we decided to format specs without nesting now?

plugins/modules/ndo_pod_settings.py
```diff
@@ -0,0 +1,304 @@
...
+    description:
+    - The UUID of the Pod Settings.
+    - This parameter is required when the O(name) needs to be updated.
+    type: str
+    aliases: [ pod_settings_uuid ]
+  description:
+    description:
+    - The description of the Pod Settings.
+    type: str
+  macsec_policy:
```
samiib, line 51: Should `macsec_policy` and `ntp_policy` be in a dictionary with a `name` option? This would be consistent with other modules.

plugins/modules/ndo_pod_profile.py
```diff
@@ -0,0 +1,357 @@
...
+
+        # Retrieve a summary of all fabric_policy templates from NDO and loop through all templates to find the pod settings policy
+        for template in MSOTemplate(mso, "fabric_policy").template:
+            for policy in template.get("policies", []):
+                # Only retrieve the template details if the fabric_policy template has a pod settings policy defined
+                if policy.get("objType") == "podPolicyGroup" and policy.get("count") > 0:
+                    pod_settings_template = mso_templates.get_template("fabric_policy", template.get("templateName"), template.get("templateId"))
+                    match = pod_settings_template.get_pod_settings_object(pod_settings_uuid)
+                    if match:
+                        pod_profile["policyName"] = match.details.get("name")
```
samiib, line 351: Should the pod settings templateId and templateName be added?

plugins/modules/ndo_pod_profile.py
```diff
@@ -0,0 +1,357 @@
...
+            mso.existing = match.details  # When the state is present
+        else:
+            mso.existing = {}  # When the state is absent
+    elif module.check_mode and state != "query":  # When the state is present/absent with check mode
+        set_pod_profile_policy_names(mso, mso_templates, mso.proposed)
+        mso.existing = mso.proposed if state == "present" else {}
+    mso.exit_json()
+
+
+def set_pod_profile_policy_names(mso, mso_templates, pod_profile):
```
samiib, line 315: Should this function also be adding the templateName/templateId of the pod profile?

samiib (changes requested):

plugins/module_utils/template.py
akinross: The UI names it in plural form and so does the template, so I would say it would be in plural form. If you think this should be done differently let me know.

plugins/modules/ndo_pod_profile.py
akinross: I do not know all the constant changes that have decided but if you can provide me this then I can change. I thought at least it was like this in the past that for required attributes we would flatten the structure, because the user input would be easier.

plugins/modules/ndo_pod_profile.py
akinross: will add

plugins/modules/ndo_pod_profile.py
akinross: will add

plugins/modules/ndo_pod_settings.py
akinross: I changed the value to macsec_policy_name and ntp_policy_name. This way we can still use macsec_policy and ntp_policy for when configuration from different template will be allowed at a later stage. Is this something that would work for you?

shrsr (dismissed):
LGTM

gmicol (dismissed):
LGTM

plugins/modules/ndo_pod_profile.py
samiib: Thanks for changing it. I believe this was decided in a standup or PR conversation before we improved our process to document these types of decisions shared in the stand-up channel. 
Some of the most recent module in ansible-mso are doing this now. eg [(1)](https://github.com/CiscoDevNet/ansible-mso/pull/632/files#diff-972541f30f207ab1daddbdf780f43a453bbff2e70c7b753e99840782583f64d5R71) [(2)](https://github.com/CiscoDevNet/ansible-mso/pull/656/files#diff-6c3da0f6aab6fe2dfd4d8115ada8847a743d15a8acf9923b71d22d6c1846d5cdR74) [(3)](https://github.com/CiscoDevNet/ansible-mso/pull/639/files#diff-0ad43095b91619161276e948dbc4ef0c260c82991ab9b82e1250ba9974a72345R79)

plugins/modules/ndo_pod_settings.py
samiib: I also had the same thought. Although there was concern that we would need to adjust the options later if some object references could be from a different template. The team decided on a dictionary with only `name` in case `template` needed to be added later. If we go with prefixing `_name` instead, we may still have to handle backwards compatibility in some way with a dictionary and the `_name` option together. 
I don't mind either way. However, @anvitha-jain are you able to confirm the discussion from standup. This was little earlier around end of January.

akinross: Just let me know what you want and the team decides. I do not see any value of the dict approach.

plugins/module_utils/template.py
akinross: ok changed the comment above

plugins/modules/ndo_pod_settings.py
akinross: I do not agree with the solution and think `_name` is preferred since the dictionary only adds increased complexity, but I guess this is preferred. Have changed to use a dictionary to avoid any further discussion on this topic.

anvitha-jain (approved):
LGTM

gmicol (approved):
LGTM

sajagana (approved):
LGTM!

samiib (approved):
LGTM

PR #673, opened 2025-05-21
[minor_change] Added l3out_uuid and l3out template object reference support to the mso_schema_site_external_epg module

sajagana:
Note:

Formatted mso_schema_site_external_epg module test file and added additional tasks for the NDO L3Out object reference support.
PR is part of the [DCNE-376](https://cisco-aci.atlassian.net/browse/DCNE-376)


plugins/modules/mso_schema_site_external_epg.py
```diff
@@ -166,20 +192,25 @@ def main():
     l3out = module.params.get("l3out")
     l3out_schema = module.params.get("l3out_schema")
     l3out_template = module.params.get("l3out_template")
+    l3out_template_id = module.params.get("l3out_template_id")
     l3out_on_apic = module.params.get("l3out_on_apic")
     route_reachability = module.params.get("route_reachability")
+    l3out_uuid = module.params.get("l3out_uuid")
+    l3out_on_l3out_template = module.params.get("l3out_on_l3out_template")
```
akinross, line 199: do we need this option, or can we do lookup and when l3out template not found in schema it does a lookup in templates?

plugins/modules/mso_schema_site_external_epg.py
```diff
@@ -166,20 +192,25 @@ def main():
...
+    if not l3out_on_l3out_template:
+        l3out_template = template if l3out_template is None else l3out_template.replace(" ", "")
+        l3out_schema = schema if l3out_schema is None else l3out_schema
+        l3out_schema_id = mso.lookup_schema(l3out_schema)
 
     mso_schema = MSOSchema(mso, schema, template, site)
     mso_objects = mso_schema.schema_objects
 
     mso_schema.set_template_external_epg(external_epg, fail_module=False)
+    mso_template = MSOTemplate(mso, "l3out", l3out_template, l3out_template_id)
```
akinross, line 213: do we need to create this for scenarios where it is not l3out_on_l3out_template?

plugins/modules/mso_schema_site_external_epg.py
```diff
@@ -141,10 +163,13 @@ def main():
         schema=dict(type="str", required=True),
         template=dict(type="str", required=True),
         site=dict(type="str", required=True),
+        l3out_uuid=dict(type="str"),
         l3out=dict(type="str", aliases=["l3out_name"]),
         l3out_schema=dict(type="str"),
         l3out_template=dict(type="str"),
+        l3out_template_id=dict(type="str"),
         l3out_on_apic=dict(type="bool"),
+        l3out_on_l3out_template=dict(type="bool"),
```
akinross, line 172: what happens when l3out_on_apic and l3out_on_l3out_template are both set to true? should we prevent this to be allowed?

akinross (changes requested):

PR #676, opened 2025-05-28
[minor_change] Add module ndo_l3out_routed_interface to configure nodes, routed port and port-channel interfaces in l3out templates (DCNE-428)

akinross:
fixes DCNE-428

part of #627

### DIFF
--- a/plugins/module_utils/l3out_node.py
+++ b/plugins/module_utils/l3out_node.py
@@ -1,0 +1,56 @@
+# -*- coding: utf-8 -*-
+
+# Copyright: (c) 2025, Akini Ross (@akinross) <akinross@cisco.com>
+# GNU General Public License v3.0+ (see LICENSE or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+
+__metaclass__ = type
+
+from ansible_collections.cisco.mso.plugins.module_utils.utils import append_update_ops_data, delete_none_values
+
+
+class L3OutNode:
+    def __init__(self, details, mso_l3out_template, l3out_object, pod_id, node_id):
+        self.pod_id = pod_id
+        self.node_id = node_id
+        self.node_router_id = details.get("node_router_id")
+        self.node_group_policy = details.get("node_group_policy")
+        self.use_router_id_as_loopback = details.get("use_router_id_as_loopback")
+        self.node_loopback_ip = details.get("node_loopback_ip")
+        self.node = mso_l3out_template.get_l3out_nodes(l3out_object.details, self.pod_id, self.node_id)
+        self.path = "/l3outTemplate/l3outs/{0}/nodes/{1}".format(l3out_object.index, self.node.index if self.node else "-")
+
+    def construct_node_payload(self):
+        return delete_none_values(
+            {
+                "group": self.node_group_policy,
+                "podID": self.pod_id,
+                "nodeID": self.node_id,
+                "routerID": self.node_router_id,
+                "useRouteIDAsLoopback": self.use_router_id_as_loopback,
+                "loopbackIPs": [self.node_loopback_ip] if self.node_loopback_ip else None,
+            }
+        )
+
+    def update_ops(self, ops):
+        if self.node:
+            self.set_node_replace_ops(ops)
+        else:
+            self.set_node_add_op(ops)
+
+    def set_node_replace_ops(self, ops):
+        remove_data = []
+        node_payload = self.construct_node_payload()
+        if node_payload.get("useRouteIDAsLoopback") is True or node_payload.get("loopbackIPs") == [""]:
+            remove_data.append("loopbackIPs")
+        append_update_ops_data(ops, self.node.details, self.path, node_payload, remove_data)
+
+    def set_node_add_op(self, ops):
+        ops.append(self.get_node_add_op())
+
+    def get_node_add_op(self):
+        return {"op": "add", "path": self.path, "value": delete_none_values(self.construct_node_payload())}
+
+    def get_node_remove_op(self):
+        return {"op": "remove", "path": self.path}
--- a/plugins/modules/ndo_l3out_routed_interface.py
+++ b/plugins/modules/ndo_l3out_routed_interface.py
@@ -1,0 +1,544 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# Copyright: (c) 2025, Akini Ross (@akinross) <akinross@cisco.com>
+
+# GNU General Public License v3.0+ (see LICENSE or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+
+from __future__ import absolute_import, division, print_function
+
+__metaclass__ = type
+
+ANSIBLE_METADATA = {"metadata_version": "1.1", "status": ["preview"], "supported_by": "community"}
+
+DOCUMENTATION = r"""
+---
+module: ndo_l3out_routed_interface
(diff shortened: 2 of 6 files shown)


sajagana (changes requested):

anvitha-jain (approved):
LGTM

gmicol (approved):
LGTM

plugins/modules/ndo_l3out_routed_interface.py
```diff
@@ -0,0 +1,544 @@
...
+- The O(l3out) or O(l3out_uuid) must exist before using this module in your playbook.
+  The M(cisco.mso.ndo_l3out_template) module can be used for this.
+seealso:
+- module: cisco.mso.ndo_template
+- module: cisco.mso.ndo_l3out_template
+extends_documentation_fragment: cisco.mso.modules
+"""
+
+EXAMPLES = r"""
+- name: Create a L3Out Routed Interface of type port
```
gmicol, line 210:
```suggestion
- name: Create a L3Out Routed Interface of type port and a L3Out Node attached to it
```
something like that to be clearer

plugins/modules/ndo_l3out_routed_interface.py
```diff
@@ -0,0 +1,544 @@
...
+    mtu: inherit
+    node_id: 101
+    path: eth1/1
+    interface_group_policy: interface_group_policy_name
+    node_router_id: 1.1.1.1
+    node_group_policy: node_group_policy_name
+    use_router_id_as_loopback: true
+    state: present
+
+- name: Create a L3Out Routed Interface of type port_channel with reference
```
gmicol, line 231:
```suggestion
- name: Create a L3Out Routed Interface of type port_channel with reference and a L3Out Node attached to it
```
something like that to be clearer

plugins/modules/ndo_l3out_routed_interface.py
```diff
@@ -0,0 +1,544 @@
...
+  The M(cisco.mso.ndo_template) module can be used for this.
+- The O(l3out) or O(l3out_uuid) must exist before using this module in your playbook.
+  The M(cisco.mso.ndo_l3out_template) module can be used for this.
+seealso:
+- module: cisco.mso.ndo_template
+- module: cisco.mso.ndo_l3out_template
+extends_documentation_fragment: cisco.mso.modules
+"""
+
+EXAMPLES = r"""
```
gmicol, line 209: Since this module is unusual as you can also manage L3Out Nodes, In the following example section, shouldn't we give 4 examples of creating, updating, querying and deleting simply a L3Out Nodes attached to an existing L3Out routed interface? I know you have examples where you create L3Out nodes but while creating or updating a L3Out routed interface with full configuration.
It's an open discussion to up to you

plugins/module_utils/l3out_node.py
```diff
@@ -0,0 +1,56 @@
...
+        node_payload = self.construct_node_payload()
+        if node_payload.get("useRouteIDAsLoopback") is True or node_payload.get("loopbackIPs") == [""]:
+            remove_data.append("loopbackIPs")
+        append_update_ops_data(ops, self.node.details, self.path, node_payload, remove_data)
+
+    def set_node_add_op(self, ops):
+        ops.append(self.get_node_add_op())
+
+    def get_node_add_op(self):
+        return {"op": "add", "path": self.path, "value": self.construct_node_payload()}
```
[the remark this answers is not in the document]
akinross: I did it this for readability and decoupling. Future use which I do not know at the moment where I would just require the payload instead of the update the ops. What is your concern is with small functions for a specific task?

If you do not agree with the function I can remove it, but in that case I think if we would require further restructuring.

1. Setting the remove ops. Question would then be how to deal with the removal setting, because this would include more complex logic on the update_ops() with a removal or separate method. 
2. Should we then mark everything besides update_ops() as private methods?
3. Should we even keep any of the the separate methods for setting and getting the ops? Thus creating one single larger more complex update_ops)

plugins/modules/ndo_l3out_routed_interface.py
akinross: I am not sure if I follow what you mean. The current examples show how you can configure or remove an interface of this type ( routed ).

Interfaces ( any type ) cannot exist without the configuration of a node. Because we cannot patch them individually the node configuration is provided when it not exists. Where NDO will provide an error accordingly to what is missing on node configuration. As far as I understand the agreement made before, we do catch these errors with configuration specific dependencies but will rely for these type of situation on the response of NDO.

Furthermore the deletion of node is not something the user controls, this will automatically occur when there are no more interfaces configured for a node. Thus the retry mechanism on for the specific error in l3out_interface_request.

What examples would you like to have added? 

Perhaps we need to adjust the below documentation for the node_id attribute to make it more clear?

```
- The ID of the node (border leaf switch) where to deploy the L3Out routing protocol and node-level protocol configurations.
    - This parameter is required when O(path) is specified.
    - The node configuration is created under the L3Out template when it does not exist.
    - The node configuration is updated under the L3Out template when it already exists.
    - The node configuration is deleted under the L3Out template when there are no interfaces referencing it.

File: plugins/modules/ndo_l3out_routed_interface.py (hunk 1 above)
Comment (akinross) [reply plugins/modules/ndo_l3out_routed_interface.py]: Let me know if this is more clear for you. I chose other wording because the L3Out node is not attached to the routed interface. It is configuration that needs to be in place in order for a interface to be configured.

File: plugins/module_utils/l3out_node.py (hunk 4 above)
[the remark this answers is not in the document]
Comment (gmicol) [reply plugins/module_utils/l3out_node.py:53]: It was just to understand your logic here, but so far, that makes sense to have it like that as a prevention in possible future scenarios. it was more a question than a requested change. Thank you

File: plugins/modules/ndo_l3out_routed_interface.py (hunk 1 above)
Comment (gmicol) [reply plugins/modules/ndo_l3out_routed_interface.py]: your wording is better, that's definitely clearer, thanks

**Review by gmicol — DISMISSED** [review]
LGTM

**Review by anvitha-jain — APPROVED** [review]
LGTM

File: plugins/modules/ndo_l3out_routed_interface.py [hunk 5]
```diff
@@ -0,0 +1,544 @@
...
+    - Use C(query) to list the resource.
+    - Use C(present) to create or update the resource.
+    type: str
+    choices: [ absent, query, present ]
+    default: query
+notes:
+- The O(template) or O(template_id) must exist before using this module in your playbook.
+  The M(cisco.mso.ndo_template) module can be used for this.
+- The O(l3out) or O(l3out_uuid) must exist before using this module in your playbook.
+  The M(cisco.mso.ndo_l3out_template) module can be used for this.
```
akinross, line 202: check other required modules and also add to see also

plugins/modules/ndo_l3out_routed_interface.py
```diff
@@ -0,0 +1,544 @@
...
+  node_loopback_ip:
+    description:
+    - The loopback IP address of the node.
+    type: str
+    aliases: [ loopback_ip ]
+  path:
+    description:
+    - The path of the interface.
+    - The path is an existing physical port (eth1/1).
+    - This parameter or O(port_channel) is required.
```
akinross, line 80: expand description for path/portchannel: https://github.com/CiscoDevNet/ansible-mso/pull/677/files#diff-3ac79155bcb61ba026579b561a90c07f4fe83063df886e899ce599d7be557130R119

samiib (approved):
LGTM

gmicol (approved):
LGTM

sajagana (approved):
LGTM!

lhercot (approved):
LGTM

PR #698, opened 2025-07-31
Changed the update logic of the mso_schema_template_bd and mso_schema_template_vrf_rp modules and added additional tests (DCNE-304 DCNE-305)

sajagana:
Note:

Reformatted mso_schema_template_bd and mso_schema_template_vrf_rp module test files

### DIFF
--- a/plugins/module_utils/utils.py
+++ b/plugins/module_utils/utils.py
@@ -91,7 +91,7 @@ def append_update_ops_data(ops, existing_data, update_path, replace_data=None, r
     ]
     """
 
-    def recursive_replace(data, path, keys, new_value):
+    def recursive_add_replace(data, path, keys, new_value):
         key = keys[0]
         if len(keys) == 1:
             # Update the existing configuration
@@ -106,7 +106,7 @@ def append_update_ops_data(ops, existing_data, update_path, replace_data=None, r
                     )
                 )
         elif key in data:
-            recursive_replace(data[key], "{}/{}".format(path, key), keys[1:], new_value)
+            recursive_add_replace(data[key], "{}/{}".format(path, key), keys[1:], new_value)
 
     def recursive_delete(data, path, keys):
         key = keys[0]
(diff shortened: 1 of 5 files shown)


plugins/module_utils/utils.py
```diff
@@ -96,10 +96,11 @@ def recursive_replace(data, path, keys, new_value):
         if len(keys) == 1:
             # Update the existing configuration
             if new_value is not None and data.get(key) != new_value:
```
shrsr, line 98: Shall we change the name of this function to recursive_add_repalce() or something more appropriate?

plugins/module_utils/utils.py
```diff
@@ -96,10 +96,11 @@ def recursive_replace(data, path, keys, new_value):
...
+                operation = "replace" if data.get(key) is not None else "add"
```
shrsr, line 99: Did we test all the modules utilizing this helper function just to make sure that nothing is breaking?

shrsr (changes requested):

plugins/module_utils/utils.py
sajagana: I tested the change using the latest NDO modules, and all tests passed without any issues.

shrsr (approved):
LGTM

plugins/modules/mso_schema_template_vrf_rp.py
```diff
@@ -193,7 +193,13 @@ def main():
             if rp_type != mso.existing.get("rpType"):
                 ops.append(dict(op="replace", path=rp_path + "/rpType", value=rp_type))
             if multicast_route_map_policy and payload.get("mcastRtMapPolicyRef") != mso.existing.get("mcastRtMapPolicyRef"):
-                ops.append(dict(op="replace", path=rp_path + "/mcastRtMapPolicyRef", value=payload.get("mcastRtMapPolicyRef")))
+                ops.append(
+                    dict(
+                        op="replace" if mso.existing.get("mcastRtMapPolicyRef") else "add",
```
lhercot, line 198: Shouldn't this condition be if x is not None else?

lhercot (changes requested):

plugins/modules/mso_schema_template_vrf_rp.py
sajagana: Done!

shrsr (approved):
LGTM

anvitha-jain (approved):
LGTM

lhercot (approved):
LGTM

PR #699, opened 2025-08-20
[minor_change] Add support for Floating SVI Interface configuration in l3out templates with the ndo_l3out_floating_svi_interface module (DCNE-431)

akinross:
fixes DCNE-431

part of #627

### DIFF
--- a/plugins/modules/ndo_l3out_floating_svi_interface.py
+++ b/plugins/modules/ndo_l3out_floating_svi_interface.py
@@ -217,7 +217,6 @@ EXAMPLES = r"""
     auto_state: enabled
     state: present
 
-
 - name: Update a L3Out Floating SVI Interface and L3Out Node configuration
   cisco.mso.ndo_l3out_floating_svi_interface:
     host: mso_host


plugins/modules/ndo_l3out_floating_svi_interface.py
```diff
@@ -0,0 +1,473 @@
...
+    ipv6_link_local_address = mso.params.get("ipv6_link_local_address")
+    ipv6_dad = mso.params.get("ipv6_dad")
+    mac = mso.params.get("mac")
+    mtu = mso.params.get("mtu")
+    target_dscp = mso.params.get("target_dscp")
+    encapsulation_scope = mso.params.get("encapsulation_scope")
+    encapsulation_type = mso.params.get("encapsulation_type")
+    encapsulation_value = mso.params.get("encapsulation_value")
+    encap = None
+    if encapsulation_type and encapsulation_value:
```
samiib, line 364: Is it possible to specify only encap type without value or value without type when node_id isn't used?
Is that a valid scenario? if not maybe `required_together` can be used?

samiib (changes requested):

plugins/modules/ndo_l3out_floating_svi_interface.py
akinross: The encap input ( combination of type and value ) is currently used for the lookup of a specific item.

I am not sure what the benefit is of required_together over required_by? Why would you want this changed? 

Encap as a whole is an identifier field that is required when a present/absent state is used. When using required together on a query state without node_id the encapsulation_type and encapsulation_value are not used. If you would provide encapsulation_type or encapsulation_value it would not be required to fail, it would simply ignore that value and return all.

For creation or deletion of new floating svi both are required, to create the correct encap value. This value would never be updated since it is an identifier of the floating svi. Thus when the encapsulation_value is changed, the combination with encapsulation_type it will refer to a new entry ( or another existing entry ). Furthermore. the encapsulation_type on it's own would not be something that is possible to update because the allowed integer range for vxlan and vlan are different, which would result in an NDO error.

anvitha-jain (approved):
LGTM

plugins/modules/ndo_l3out_floating_svi_interface.py
akinross: Required by handles this case for present and absent flow, see test cases "Create a L3out floating svi interface without encapsulation_type" and "Create a L3out floating svi interface without encapsulation_value". 

For the query it is ignored when node_id is not specified, and query all is done. If you think that should not be done I can enforce required together for all three attributes so that query would fail in that case. This mean changing the behaviour on the already released modules.

samiib: Sounds good, no need to change anything if the spec already handles that case.

gmicol (approved):
LGTM

PR #703, opened 2025-09-01
[minor_change] Add support for Path Attributes in Floating SVI Interface configuration in l3out templates with the ndo_l3out_floating_svi_interface_path_attributes module (DCNE-435)

akinross:
fixes DCNE-435
part of #627


sajagana (approved):
LGTM!

anvitha-jain (dismissed):
LGTM

shrsr (dismissed):
LGTM

gmicol (dismissed):
LGTM

anvitha-jain (approved):
LGTM

gmicol (approved):
LGTM

plugins/module_utils/constants.py
```diff
@@ -19,6 +19,7 @@
 YES_OR_NO_TO_BOOL_STRING_MAP = {"yes": "true", "no": "false", True: "yes", False: "no"}
 
 ENABLED_OR_DISABLED_TO_BOOL_STRING_MAP = {"enabled": True, "disabled": False}
+BOOL_TO_ENABLED_OR_DISABLED_STRING_MAP = {True: "enabled", False: "disabled"}
```
lhercot, line 22: Any reason not to merge the two like for YES_OR_NO_TO_BOOL_STRING_MAP ?

plugins/modules/ndo_l3out_floating_svi_interface_path_attributes.py
```diff
@@ -0,0 +1,375 @@
...
+    description:
+    - The Anchor Node ID of the L3Out Floating SVI Interface.
+    type: str
+    required: true
+    aliases: [ node, anchor_node_id ]
+  encapsulation_type:
+    description:
+    - The encapsulation type of the interface.
+    type: str
+    required: true
```
lhercot, line 59: If we are making this always required I would default it to VLAN as this is the main use case.

plugins/modules/ndo_l3out_floating_svi_interface_path_attributes.py
```diff
@@ -0,0 +1,375 @@
...
+- name: Query all existing Path Attributes of L3Out Floating SVI Interface
+  cisco.mso.ndo_l3out_floating_svi_interface_path_attributes:
+    host: mso_host
+    username: admin
+    password: SomeSecretPassword
+    template: l3out_template
+    l3out: l3out_name
+    node_id: 101
+    encapsulation_type: vlan
+    encapsulation_value: 1200
```
lhercot, line 205: are those really required for query?

lhercot (changes requested):

plugins/modules/ndo_l3out_floating_svi_interface_path_attributes.py
akinross: always thought we avoided defaults, but can agree with this one

plugins/modules/ndo_l3out_floating_svi_interface_path_attributes.py
akinross: yes because the encap combination (type/value) is used while identifying the floating svi ( parent object ) of path attributes

sajagana (changes requested):

samiib (approved):
LGTM

shrsr (approved):
LGTM

anvitha-jain (approved):
LGTM

PR #704, opened 2025-09-02
Bugfix SVI interface (DCNE-538)

akinross:


lhercot (approved):
LGTM with expectation you have tested successfully all modules which uses get_site_interface_details and get_l3out_svi_interface directly or indirectly.

PR #720, opened 2025-11-04
Added Route Map Policy for Route Control and context modules (DCNE-125, DCNE-259, DCNE-260)

sajagana:


shrsr (approved):
LGTM

plugins/modules/ndo_route_map_policy_route_control_context.py
```diff
@@ -0,0 +1,434 @@
...
+  name:
+    description:
+    - The name of the Route Control context.
+    type: str
+  description:
+    description:
+    - The description of the Route Control context.
+    - Providing an empty string O(description="") will remove description from the Route Control context.
+    type: str
+  order:
```
samiib: Does order have a default or is it required? It looks required in the UI and also the UI says 0-9?

tests/integration/targets/ndo_route_map_policy_route_control/tasks/main.yml
```diff
@@ -0,0 +1,380 @@
...
+    - name: Update Route Map Policy for Route Control 1 using name
+      cisco.mso.ndo_route_map_policy_route_control:
+        <<: *mso_info
+        template: ansible_tenant_template
+        name: route_map_policy_1_updated
+        description: updated_again
+        match_type: routing_policy_only
+        state: present
+      register: update_route_map_policy_with_name
+
```
samiib: Can you add a test case that verifies all the removable properties gets removed from object?

tests/integration/targets/ndo_route_map_policy_route_control_context/tasks/main.yml
```diff
@@ -0,0 +1,566 @@
...
+          - update_route_map_policy_context_2.previous.setRuleName == "ansible_set_rule_policy_2"
+          - update_route_map_policy_context_2.previous.setRuleRef is defined
+          - update_route_map_policy_context_2.previous.setRuleTemplateId is defined
+          - update_route_map_policy_context_2.previous.setRuleTemplateName == "ansible_tenant_template"
+          - update_route_map_policy_context_2.proposed.action == "permit"
+          - rm_route_map_policy_context_2_values is changed
+          - rm_route_map_policy_context_2_values.current.action == "permit"
+          - rm_route_map_policy_context_2_values.current.matchRules == []
+          - rm_route_map_policy_context_2_values.current.name == "route_map_policy_context_2"
+          - rm_route_map_policy_context_2_values.current.order == 1
```
samiib: Looks like some of the removed properties are missed. Should setRuleName/Ref be undefined?
Also should description be undefined if it gets removed with and empty string?

plugins/modules/ndo_route_map_policy_route_control_context.py
sajagana: Updated! Also order is marked as a required field in the UI, but it defaults to 0 when we create the object using API.

tests/integration/targets/ndo_route_map_policy_route_control_context/tasks/main.yml
sajagana: Done

plugins/modules/ndo_route_map_policy_route_control_context.py
```diff
@@ -0,0 +1,441 @@
...
+      template:
+        description:
+        - The template associated with the Set Rule Policy.
+        - This parameter or O(set_rule.template_id) is required.
+        type: str
+      template_id:
+        description:
+        - The template ID associated with the Set Rule Policy.
+        - This parameter or O(set_rule.template) is required.
+        type: str
```
akinross: why is this not a uuid and reference as in match_rules? i thought that was the current approach of uuid and reference passing?

sajagana: Yes, changed!

gmicol (changes requested):

shrsr (approved):
LGTM

shrsr (approved):
LGTM

akinross (approved):
LGTM

anvitha-jain (approved):
LGTM

PR #751, opened 2026-07-10
Fix idempotency in mso_schema_template_l3out and mso_schema_template_external_epg (DCNE-868)

### LINKED ISSUE #750: mso_schema_template_l3out / mso_schema_template_external_epg are never idempotent (ref shape mismatch in previous/existing compare) (DCNE-868)
### Summary

Both `mso_schema_template_l3out` and `mso_schema_template_external_epg` report `changed=true` on every re-run against an already-converged configuration, because the reference fields compared between the "previous"/"existing" state and the "proposed"/"sent" state end up in different shapes (string form vs. dict form).

### `mso_schema_template_l3out`

At the end of `main()`, the module builds the desired L3out payload with `vrf_ref = mso.make_reference(...)`, which produces a dict shaped like `{"vrfName": ..., "schemaId": ..., "templateName": ...}`. This flows into `mso.proposed` via `mso.sanitize()`.

`mso.previous`, however, is captured directly from the schema object returned by the API (`mso.previous = mso.existing`, set before the state branch), where `vrfRef` is still the raw string-form reference (e.g. `/schemas/.../templates/.../vrfs/...`). The two representations are never reconciled before the `changed` comparison (`self.previous != self.existing` in `exit_json`), so they never compare equal, even when nothing actually changed.

Additionally, the module issues its `PATCH` request unconditionally (`if not module.check_mode:`) rather than gating on whether `mso.proposed` actually differs from `mso.previous`, so a redundant no-op PATCH is sent to the API on every run regardless.

`mso_schema_template_bd.py` already implements the correct pattern for this exact situation — it normalizes `mso.previous["vrfRef"]` via `mso.vrf_dict_from_ref()` right after `mso.existing = mso.proposed`, and gates its PATCH request on `mso.proposed != mso.previous`.

### `mso_schema_template_external_epg`

Top-level references (`vrfRef`, `l3outRef`, `anpRef`) on `mso.existing` are correctly normalized via `mso.dict_from_ref()` right when `mso.existing` is first populated. However, nested `contractRelationships[].contractRef` entries are normalized via `mso.dict_from_ref()` only on `mso.sent` (when building the update payload for an existing external EPG) — the same normalization is never applied to `mso.existing`/`mso.previous`. Once any contract relationship is bound to the external EPG, every subsequent idempotent re-run reports `changed=true` because `contractRef` stays string-form on one side of the comparison and dict-form on the other.

### Reproduction recipe

1. Create an L3out via `mso_schema_template_l3out` with a `vrf` reference, or create an external EPG via `mso_schema_template_external_epg` with a bound contract relationship.
2. Re-run the exact same task with identical parameters against the already-converged state.
3. Observe `changed: true` on the second run, when it should be `changed: false`.

### Version

Collection currently at `version: 2.13.0` per `galaxy.yml` (released 2026-03-07 per `CHANGELOG.rst`). This reproduces against the current `develop` branch head.

dtzp555-max:
### Summary

Fixes #750

Two modules were reporting `changed=true` on every idempotent re-run due to a shape mismatch between the reference fields on the "previous"/"existing" side of the comparison and the "proposed"/"sent" side.

**`mso_schema_template_l3out.py`**
- Normalizes `mso.previous["vrfRef"]` from its raw string form (as returned by the API) into the same dict shape produced by `mso.make_reference()`, right after `mso.existing = mso.proposed` is set.
- Gates the module's `PATCH` request on `mso.proposed != mso.previous` instead of issuing it unconditionally on every run, so a redundant no-op PATCH is no longer sent when nothing changed. Delete/absent-state behavior is unaffected — `mso.proposed` remains empty (`{}`) whenever there's an existing L3out to remove, so the diff check still evaluates true and the delete still fires.

**`mso_schema_template_external_epg.py`**
- Adds a normalization loop over `mso.existing`'s `contractRelationships[].contractRef` entries, mirroring the existing loop that already normalizes the same field on `mso.sent`. Previously, once any contract was bound to an external EPG, every re-run reported `changed=true` because `contractRef` stayed string-form on the existing/previous side.

Both fixes mirror the normalization pattern already established in `mso_schema_template_bd.py`, which handles the equivalent `vrfRef` comparison correctly today.

### Testing

Validated in a lab NDO environment with idempotent re-run tests.

### Changelog

Added `changelogs/fragments/fix-template-l3out-extepg-idempotency.yaml` per the collection's changelog fragment convention.

### DIFF
--- a/changelogs/fragments/fix-template-l3out-extepg-idempotency.yaml
+++ b/changelogs/fragments/fix-template-l3out-extepg-idempotency.yaml
@@ -1,0 +1,4 @@
+---
+bugfixes:
+  - Fix to avoid reporting changed on every idempotent run due to a VRF reference format mismatch between the current and desired state in mso_schema_template_l3out
+  - Fix to avoid reporting changed on every idempotent run due to a contract reference format mismatch between the current and desired state in mso_schema_template_external_epg
--- a/changelogs/fragments/fix-template-l3out-extepg-idempotency.yaml
+++ b/changelogs/fragments/fix-template-l3out-extepg-idempotency.yaml
@@ -1,4 +1,0 @@
----
-bugfixes:
-  - Fix to avoid reporting changed on every idempotent run due to a VRF reference format mismatch between the current and desired state in mso_schema_template_l3out
-  - Fix to avoid reporting changed on every idempotent run due to a contract reference format mismatch between the current and desired state in mso_schema_template_external_epg
--- a/plugins/modules/mso_schema_template_external_epg.py
+++ b/plugins/modules/mso_schema_template_external_epg.py
@@ -275,6 +275,9 @@ def main():
             mso.existing["l3outRef"] = mso.dict_from_ref(mso.existing.get("l3outRef"))
         if "anpRef" in mso.existing:
             mso.existing["anpRef"] = mso.dict_from_ref(mso.existing.get("anpRef"))
+        for contract in mso.existing.get("contractRelationships", []) or []:
+            if contract.get("contractRef") and not isinstance(contract.get("contractRef"), dict):
+                contract["contractRef"] = mso.dict_from_ref(contract.get("contractRef"))
 
     if state == "query":
         if external_epg is None:
@@ -327,9 +330,6 @@ def main():
                 del mso.existing["anpRef"]
             # clean contractRef to fix api issue
             for contract in mso.sent.get("contractRelationships"):
-                contract["contractRef"] = mso.dict_from_ref(contract.get("contractRef"))
-            # normalize contractRef on existing/previous so idempotency comparisons match mso.sent's shape
-            for contract in mso.existing.get("contractRelationships", []) or []:
                 if not isinstance(contract.get("contractRef"), dict):
                     contract["contractRef"] = mso.dict_from_ref(contract.get("contractRef"))
             ops.append(dict(op="replace", path=eepg_path, value=mso.sent))
--- a/plugins/modules/mso_schema_template_external_epg.py
+++ b/plugins/modules/mso_schema_template_external_epg.py
@@ -328,6 +328,10 @@ def main():
             # clean contractRef to fix api issue
             for contract in mso.sent.get("contractRelationships"):
                 contract["contractRef"] = mso.dict_from_ref(contract.get("contractRef"))
+            # normalize contractRef on existing/previous so idempotency comparisons match mso.sent's shape
+            for contract in mso.existing.get("contractRelationships", []) or []:
+                if not isinstance(contract.get("contractRef"), dict):
+                    contract["contractRef"] = mso.dict_from_ref(contract.get("contractRef"))
             ops.append(dict(op="replace", path=eepg_path, value=mso.sent))
         else:
             ops.append(dict(op="add", path=eepgs_path + "/-", value=mso.sent))
--- a/plugins/modules/mso_schema_template_l3out.py
+++ b/plugins/modules/mso_schema_template_l3out.py
@@ -218,8 +218,10 @@ def main():
             ops.append(dict(op="add", path=l3outs_path + "/-", value=mso.sent))
 
         mso.existing = mso.proposed
+        if "vrfRef" in mso.previous and not isinstance(mso.previous.get("vrfRef"), dict):
+            mso.previous["vrfRef"] = mso.vrf_dict_from_ref(mso.previous.get("vrfRef"))
 
-    if not module.check_mode:
+    if not module.check_mode and mso.proposed != mso.previous:
         mso.request(schema_path, method="PATCH", data=ops)
 
     mso.exit_json()
--- a/plugins/modules/mso_schema_template_l3out.py
+++ b/plugins/modules/mso_schema_template_l3out.py
@@ -177,6 +177,8 @@ def main():
     if l3out is not None and l3out in l3outs:
         l3out_idx = l3outs.index(l3out)
         mso.existing = schema_obj.get("templates")[template_idx]["intersiteL3outs"][l3out_idx]
+        if "vrfRef" in mso.existing and not isinstance(mso.existing.get("vrfRef"), dict):
+            mso.existing["vrfRef"] = mso.dict_from_ref(mso.existing.get("vrfRef"))
 
     if state == "query":
         if l3out is None:
@@ -218,8 +220,6 @@ def main():
             ops.append(dict(op="add", path=l3outs_path + "/-", value=mso.sent))
 
         mso.existing = mso.proposed
-        if "vrfRef" in mso.previous and not isinstance(mso.previous.get("vrfRef"), dict):
-            mso.previous["vrfRef"] = mso.vrf_dict_from_ref(mso.previous.get("vrfRef"))
 
     if not module.check_mode and mso.proposed != mso.previous:
         mso.request(schema_path, method="PATCH", data=ops)
--- a/tests/integration/targets/mso_schema_template_external_epg/tasks/main.yml
+++ b/tests/integration/targets/mso_schema_template_external_epg/tasks/main.yml
@@ -1095,6 +1095,13 @@
     that:
     - nm_query_epg1_contract.current.contractRelationships | length == 4
 
+- name: Verify contractRef is a dict on query result
+  ansible.builtin.assert:
+    that:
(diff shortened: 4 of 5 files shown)


plugins/modules/mso_schema_template_external_epg.py
```diff
@@ -328,6 +328,10 @@ def main():
             # clean contractRef to fix api issue
             for contract in mso.sent.get("contractRelationships"):
                 contract["contractRef"] = mso.dict_from_ref(contract.get("contractRef"))
+            # normalize contractRef on existing/previous so idempotency comparisons match mso.sent's shape
+            for contract in mso.existing.get("contractRelationships", []) or []:
+                if not isinstance(contract.get("contractRef"), dict):
+                    contract["contractRef"] = mso.dict_from_ref(contract.get("contractRef"))
```
akinross, line 334: i think the normalisation should be done at lines https://github.com/CiscoDevNet/ansible-mso/pull/751/changes#diff-6b5534abc2a88eae4eb47ecbffcedc1a7e4f14518731811f0f2bd03e3bb72922R267-R277 so that the normalisation off existing occurs for every state and not only state=present

plugins/modules/mso_schema_template_l3out.py
```diff
@@ -218,8 +218,10 @@ def main():
             ops.append(dict(op="add", path=l3outs_path + "/-", value=mso.sent))
 
         mso.existing = mso.proposed
+        if "vrfRef" in mso.previous and not isinstance(mso.previous.get("vrfRef"), dict):
+            mso.previous["vrfRef"] = mso.vrf_dict_from_ref(mso.previous.get("vrfRef"))
```
akinross: i think it would be possible to do the normalisation at https://github.com/CiscoDevNet/ansible-mso/pull/751/changes#diff-749ca20629c7caacf162e31b9c6d0166ab275d2b8ecb755d5d13c9db2dc11fccR177-R179. similar to the external epg comment

changelogs/fragments/fix-template-l3out-extepg-idempotency.yaml
```diff
@@ -0,0 +1,4 @@
+---
+bugfixes:
```
akinross: changelogs are not added manually. could you please remove this so @samiib could handle it during release?

plugins/modules/mso_schema_template_external_epg.py
akinross, line 207: please add tests to https://github.com/CiscoDevNet/ansible-mso/blob/develop/tests/integration/targets/mso_schema_template_external_epg/tasks/main.yml that verify this change is not impacting current tests and fixes the idempotency issue

plugins/modules/mso_schema_template_l3out.py
akinross, line 132: please add tests to https://github.com/CiscoDevNet/ansible-mso/blob/develop/tests/integration/targets/mso_schema_template_l3out/tasks/main.yml that verify this change is not impacting current tests and fixes the idempotency issue

akinross (changes requested):

plugins/modules/mso_schema_template_l3out.py
dtzp555-max: Done in 86614bf — normalization now happens right after `mso.existing` is captured from the intersiteL3outs lookup (mirroring the external EPG module), and the present-branch block is removed. Switched to the general `mso.dict_from_ref()` to match the external EPG capture block; it produces the same `{vrfName, schemaId, templateName}` shape as `make_reference()`.

changelogs/fragments/fix-template-l3out-extepg-idempotency.yaml
dtzp555-max: Done in 86614bf — fragment removed.

plugins/modules/mso_schema_template_external_epg.py
dtzp555-max: Done in 86614bf — moved the contractRef normalization into the block that already normalizes vrfRef/l3outRef/anpRef when `mso.existing` is first captured, so it now applies to every state.

One knock-on effect: `sanitize()` deepcopies `mso.existing` into `mso.sent`, so on an update of an EPG that already has bound contracts, the pre-existing loop that normalizes the contractRef entries of `mso.sent` now receives already-dict values, and the unguarded `mso.dict_from_ref()` would raise a TypeError (it runs a regex on its input). Added an `isinstance` guard to that loop so string-form refs are still normalized and dict-form ones are left as-is.

plugins/modules/mso_schema_template_external_epg.py
dtzp555-max: This target already contains the exact idempotency regression test for this fix: "Add EPG 1 again (normal_mode)" re-runs the module against EPG 1 with 4 bound contracts and asserts `nm_add_epg_1_again is not changed` — an assertion that only passes with this change in place. To cover the new behavior introduced by moving the normalization to the capture point (it now applies to every state), I added an assertion in 86614bf that the `state: query` result returns `contractRef` in its normalized dict form (`contractName`/`templateName`/`schemaId`).

plugins/modules/mso_schema_template_l3out.py
dtzp555-max: Same as the external EPG target — the existing "Add L3Out again" task already asserts `add_l3out_again is not changed` against an L3Out bound to VRF1, which is the idempotency regression this fix addresses. Added in 86614bf an assertion that the `state: query` result returns `vrfRef` in its normalized dict form (`vrfName`/`templateName`/`schemaId`), covering the capture-point normalization applying to query as well.

dtzp555-max:
Review comments addressed in 86614bf:

- Normalization moved to the point where `mso.existing` is first captured in both modules, so it applies to every state (present/absent/query), and the present-branch blocks were removed.
- Changelog fragment removed.
- Added query-shape assertions to both integration targets; the pre-existing "add again → is not changed" tasks in both targets cover the idempotent re-run itself.

Re-validated the updated patch in a lab NDO environment: idempotent re-runs report `changed=false` (including an external EPG with a bound contract), `state: query` returns dict-form refs, and `state: absent` still issues the delete.

tests/integration/targets/mso_schema_template_external_epg/tasks/main.yml
```diff
@@ -1095,6 +1095,13 @@
     that:
     - nm_query_epg1_contract.current.contractRelationships | length == 4
 
+- name: Verify contractRef is a dict on query result
+  ansible.builtin.assert:
+    that:
+    - nm_query_epg1_contract.current.contractRelationships[0].contractRef.contractName is defined
+    - nm_query_epg1_contract.current.contractRelationships[0].contractRef.templateName is defined
```
akinross: is there a reason to use is defined for contractName and templateName? these should be known values which could be asserted similar to vrfName/templateName in your other test add

akinross (changes requested):

tests/integration/targets/mso_schema_template_external_epg/tasks/main.yml
dtzp555-max: No reason other than caution about relying on element order — updated in 42f9591 to assert the known values (`contractName == "Contract1"`, `templateName == "Template1"`), matching the l3out test.

akinross (approved):
assuming all the tests are passing in our environment

sajagana (approved):
LGTM!

samiib (approved):
Tests Pass ✅ 

```
mso_schema_template_l3out

PLAY RECAP *********************************************************************
ndo_3_2                    : ok=28   changed=7    unreachable=0    failed=0    skipped=1    rescued=0    ignored=4   
ndo_4_1                    : ok=28   changed=8    unreachable=0    failed=0    skipped=1    rescued=0    ignored=4   
ndo_4_2                    : ok=28   changed=8    unreachable=0    failed=0    skipped=1    rescued=0    ignored=4

mso_schema_template_external_epg

PLAY RECAP *********************************************************************
ndo_3_2                    : ok=113  changed=43   unreachable=0    failed=0    skipped=2    rescued=0    ignored=6   
ndo_4_1                    : ok=105  changed=41   unreachable=0    failed=0    skipped=10   rescued=0    ignored=6   
ndo_4_2                    : ok=105  changed=41   unreachable=0    failed=0    skipped=10   rescued=0    ignored=6
```

PR #755, opened 2026-08-31
Deprecate mso schema template anp epg selector (DCNE-895)

akinross:
DCNE-895

### DIFF
--- a/meta/runtime.yml
+++ b/meta/runtime.yml
@@ -18,7 +18,7 @@ plugin_routing:
           has been released for ND v2.2 (NDO v4.1) and later.
     mso_schema_template_anp_epg_selector:
       deprecation:
-        removal_version: 3.0.0
+        removal_version: 4.0.0
         warning_text: This module manages cloud-specific features that are no longer supported in Nexus Dashboard
           4.x (NDO 5.x) releases.
 action_groups:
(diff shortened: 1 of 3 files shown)


sajagana (dismissed):
LGTM!

sajagana (dismissed):
LGTM!

meta/runtime.yml
```diff
@@ -16,6 +16,11 @@ plugin_routing:
         removal_version: 3.0.0
         warning_text: Due to changes in the ND and NDO API, a new module (cisco.mso.ndo_schema_template_deploy)
           has been released for ND v2.2 (NDO v4.1) and later.
+    mso_schema_template_anp_epg_selector:
+      deprecation:
+        removal_version: 3.0.0
```
samiib: Should this be 4.0.0 instead? The next release version which will include this new deprecation would most likely be 3.0.0.

samiib (changes requested):

meta/runtime.yml
akinross: If you think this is better can do, but did not see any breaking changes for now thus kept it on 3.0.0. Also since it is cloud which is already not supported platform anymore thought we might as well do this in 3.0.0 release

samiib (approved):

sajagana (approved):
LGTM!

PR #758, opened 2026-09-03
Deprecate dhcp policy and relay modules (DCNE-904, DCNE-914, DCNE-929)

akinross:
DCNE-904
DCNE-914
DCNE-929

### DIFF
--- a/meta/runtime.yml
+++ b/meta/runtime.yml
@@ -65,28 +65,28 @@ plugin_routing:
       deprecation:
         removal_version: 4.0.0
         warning_text: This module is no longer functional on NDO 4.0 and later. The replacement
-          cisco.mso.ndo_dhcp_option_policy module is supported on ND v3.1 (NDO v4.3) and later.
+          cisco.mso.ndo_dhcp_option_policy module is supported on ND v3.1 / NDO v4.3 and later.
     mso_dhcp_option_policy_option:
       deprecation:
         removal_version: 4.0.0
         warning_text: This module is no longer functional on NDO 4.0 and later. The replacement
-          cisco.mso.ndo_dhcp_option_policy module is supported on ND v3.1 (NDO v4.3) and later.
+          cisco.mso.ndo_dhcp_option_policy module is supported on ND v3.1 / NDO v4.3 and later.
     mso_dhcp_relay_policy:
       deprecation:
         removal_version: 4.0.0
         warning_text: This module is no longer functional on NDO 4.0 and later. The replacement
-          cisco.mso.ndo_dhcp_relay_policy module is supported on ND v3.1 (NDO v4.3) and later.
+          cisco.mso.ndo_dhcp_relay_policy module is supported on ND v3.1 / NDO v4.3 and later.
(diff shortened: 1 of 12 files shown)


sajagana (dismissed):
LGTM!

meta/runtime.yml
```diff
@@ -61,6 +61,32 @@ plugin_routing:
         removal_version: 4.0.0
         warning_text: This module manages cloud-specific features that are no longer supported in Nexus Dashboard
           4.x (NDO 5.x) releases.
+    mso_dhcp_option_policy:
+      deprecation:
+        removal_version: 4.0.0
+        warning_text: This module is no longer functional on NDO 4.0 and later. The replacement
+          cisco.mso.ndo_dhcp_option_policy module is supported on ND v3.1 (NDO v4.3) and later.
```
samiib: The other PR uses `ND v3.1 / NDO v4.3` format. Should these be consistent?

meta/runtime.yml
```diff
@@ -61,6 +61,32 @@ plugin_routing:
...
+    mso_dhcp_relay_policy_provider:
+      deprecation:
+        removal_version: 4.0.0
+        warning_text: This module is no longer functional on NDO 4.0 and later. The replacement
+          cisco.mso.ndo_dhcp_relay_policy module is supported on ND v3.1 (NDO v4.3) and later.
+    mso_schema_template_bd_dhcp_policy:
+      deprecation:
+        removal_version: 4.0.0
+        warning_text: This module is no longer functional on NDO 4.0 and later. The replacement
+          cisco.mso.ndo_schema_template_bd_dhcp_policy module requires PATCH support, so it is supported on ND v3.1
```
samiib: Do we need to specify the PATCH support reason? Or is this saying that the ndo_ module still needs patch support?

samiib (changes requested):

meta/runtime.yml
akinross: the ndo replacement module required PATCH to be available on the endpoint in order to work which got introduced in 4.3, so there is a gap in usage. i can reword if something is unclear of can remove the reason if it is confusing. let me know what you think would be more clear and i will change

samiib:
```suggestion
          cisco.mso.ndo_schema_template_bd_dhcp_policy module is supported on ND v3.1 / NDO 4.3 and later.
```

I think the same as the above module is okay. It also has a gap.

sajagana (dismissed):
LGTM!

plugins/modules/mso_schema_template_bd_dhcp_policy.py
```diff
@@ -65,6 +65,11 @@
     type: str
     choices: [ absent, present, query ]
     default: present
+deprecated:
+  removed_in: '4.0.0'
+  why: This module is no longer functional on NDO 4.0 and later. The replacement module requires PATCH support, so it
```
samiib: The same deprecation message update can be made here too.

sajagana (approved):
LGTM!

samiib (approved):
LGTM

PR #762, opened 2026-09-07
[ignore] fix ndo_tenant_igmp_interface_policy tests to avoid new multcast entries validation rule (DCNE-851)

akinross:
DCNE-851

New validation rule got introduced in ND 4.2 which fails test. Updated the config input to comply to this rule.

Invalid configuration for IGMP Interface Policy  ansible_igmp_interface_policy_2_updated: IGMP interface policy 'ansible_igmp_interface_policy_2_updated' has reserved multicast entries but no statelimit route map 
policy reference configured. Statelimit route map policy multicast reference is required when using reserved multicast entries


sajagana (approved):
LGTM!

samiib (approved):

PR #763, opened 2026-09-07
[ignore] fix mso_schema_template_anp_epg_useg_attribute tests to avoid new vrf and bd presence on epg configuration validation rule (DCNE-905)

akinross:
DCNE-905


sajagana (approved):
LGTM!

samiib (approved):

PR #765, opened 2026-09-08
[minor_change] Deprecate display_name, users and remote_users in mso_tenant on ND 4.2+ / NDO 5.2+ (, ) (DCNE-881)

### LINKED ISSUE #753: mso_tenant is unusable on ND 4.2 (DCNE-881)
### Community Note

### Description

On Nexus Dashboard (ND) 4.2+ / NDO 5.2+, every superuser is automatically and permanently
associated with a built-in "all-tenants-domain" -- as the name implies, this domain applies to
**every tenant**, without exception, not just a specific one. Users in this domain cannot be
removed from any tenant's user associations via the API. As a result, `mso_tenant` with
`state: present` fails against **any** tenant on ND 4.2+, even when the `users` parameter is
not specified at all, making the module effectively unusable on this platform version.

Root cause: in `plugins/module_utils/mso.py`, `lookup_users()` substitutes `["admin"]` when
`users` is `None`:

```python
def lookup_users(self, users, ignore_not_found_error=False):
    """Look up users and return their ids"""
    # Ensure tenant has at least admin user
    if users is None:
        users = ["admin"]
    elif "admin" not in users:
        users.append("admin")
    ...
```

`mso_tenant.py` then sends this computed list as `userAssociations`, which (per the module's
own documentation) fully replaces existing associated users rather than merging with them. On
ND 4.2+, this implicitly attempts to remove any already-associated all-tenants-domain user,
which the API rejects.

This affects any `mso_tenant` `state: present` call (not specific to any one playbook/module)
run against **any** tenant on ND 4.2+, since the all-tenants-domain applies universally, not
to a specific tenant. There is currently no way for a caller to discover which users belong to
"all-tenants-domain" ahead of time in order to include them explicitly in `users:`.

* Reformat mso_tenant/lookup_users so it does not attempt to remove all-tenants-domain users
  from a tenant's associations, since the ND 4.2+ platform does not allow this.

### Affected Module Name(s):

* mso_tenant

* V 5.2 and ND-based.

* N/A (tenant-level resource, not site-level).

### Collection versions

* cisco.mso 2.13.0

### Output/ Error message

```
Cannot remove user '<uuid>' from tenant 'ansible_test' - users associated with
all-tenants-domain are immutable.
```

### Expected Behavior

`mso_tenant` with `state: present` should succeed without attempting to remove
all-tenants-domain users from the tenant's associations -- particularly when the `users`
parameter is not specified at all, in which case no user-association changes should be
implied. Since every tenant is subject to the all-tenants-domain, this should work
unconditionally on ND 4.2+, not just for tenants without pre-existing superuser associations.

### Actual Behavior

The module always computes a `userAssociations` list (defaulting to `["admin"]` when `users`
is not specified) and sends it as a full replacement. On ND 4.2+, since every tenant already
has an all-tenants-domain superuser associated by default, this always implicitly attempts to
remove that user, and the API returns a 400 error on essentially every `state: present` call,
causing the task to fail.

### Playbook tasks to Reproduce

```yaml
- cisco.mso.mso_tenant:
    output_level: debug
    host: '{{ mso_hostname }}'
    username: '{{ mso_username }}'
    password: '{{ mso_password }}'
    validate_certs: '{{ mso_validate_certs | default(false) }}'
    name: ansible_test
    state: present
```

Run against a tenant on an ND 4.2+ / NDO 5.2+ system. Since the all-tenants-domain applies to
every tenant by default, this fails universally, not just for a specific pre-touched tenant.

### Important Factoids

* Discovered while running Netflow module (`ndo_tenant_netflow_exporter`/`_record`/`_monitor`)
  integration tests against a multi-version lab (ND 3.2, ND 4.1, ND 4.2). The failure occurred
  in shared test setup (`mso_tenant`), not in the Netflow modules themselves.
* The identical task, run against the same tenant name on an older ND/NDO version (ND 4.1) in
  the same lab, did not fail -- all-tenants-domain and its immutability restriction appear to
  be new to ND 4.2+.
* Not specific to any one module -- any playbook/test that calls `mso_tenant` with
  `state: present` on ND 4.2+ can hit this, since the all-tenants-domain is universal (applies
  to all tenants), not scoped to specific tenants.

### References

* N/A

samiib:
Fixes #753

- DCNE-881
- DCNE-920

## Summary

Fixes `mso_tenant` failures on Nexus Dashboard (ND) 4.2+ / NDO 5.2+, where `userAssociations` is now derived from tenant-domain membership and partly platform-managed/immutable. Sending a full-replace payload (as before) caused `400` errors and perpetual idempotency drift. Resolved via **deprecation** of the affected parameters, mirroring the approach already used for `mso_tenant` in `terraform-provider-mso`, rather than adding new version-specific enforcement logic.

## Changes

- **`plugins/module_utils/mso.py`**
  - Added `is_platform_version_at_least()` to compare `major.minor` platform versions (tolerant of suffixes like `5.2(1)`/`5.2a`).
  - Added `MSOModule.get_platform_version()` (cached `platform/version` lookup).
  - `lookup_users()`: no longer force-adds `"admin"` on ND 4.2+ / NDO 5.2+ (older versions unchanged). Also skips the local/remote user list fetch when there's nothing to look up.
  - `check_changed()`: `ignore_keys` now accepts dotted paths (e.g. `siteAssociations.awsAccount`) to exclude specific nested/list-item keys from the diff, in addition to whole top-level keys, without mutating `sent`/`existing`.

- **`plugins/modules/mso_tenant.py`**
  - On ND 4.2+ / NDO 5.2+, `userAssociations` is only included in the payload when the caller explicitly sets `users`/`remote_users`; otherwise the API fully owns it.
  - Emits a `module.deprecate()` warning when `users`/`remote_users` are used on ND 4.2+ / NDO 5.2+.
  - `check_changed()` ignores platform-added `userAssociations` noise and unmanaged `siteAssociations` cloud-account keys, but still detects real caller-intended user changes (e.g. adding a not-yet-associated user).
  - `DOCUMENTATION` updated for `users`, `remote_users` (deprecated on ND 4.2+/NDO 5.2+) and `display_name` (must equal `tenant` on ND 4.2+ or `sites` updates are rejected; doc-only note, no new enforcement).

- **`tests/integration/targets/mso_tenant/tasks/main.yml`**
  - Removed remote-user test coverage (no longer applicable/reliable on ND 4.2+).
  - Relaxed exact-count `userAssociations` assertions to presence checks where platform-side auto-backfill makes exact counts non-deterministic.
  - Added a version-gated block (`< 5.2`, non-admin test user) covering the legacy forced-admin behavior: add a tenant with an explicit non-admin user (expect 2 associations), then update with an empty user list (expect only the forced admin remains).

## Breaking changes

None. Behavior on  ND < 4.2 / NDO < 5.2 is unchanged. On ND 4.2+ / NDO 5.2+, `users`/`remote_users` are deprecated (warning only) rather than removed, and the module no longer force-adds `admin` there (previously partly non-functional anyway due to platform-side immutability).

## Test matrix

| Target | Platform | Result | ok | changed | failed | skipped | ignored |
|---|---|---|---|---|---|---|---|
| `ndo_3_2` | ND 3.2 | ✅ Pass | 87 | 23 | 0 | 1 | 1 |
| `ndo_4_1` | ND (< 4.2) | ✅ Pass | 87 | 23 | 0 | 1 | 1 |
| `ndo_4_2` | ND 4.2+ / NDO 5.2+ | ✅ Pass | 80 | 20 | 0 | 8 | 1 |

### DIFF
--- a/plugins/modules/mso_tenant.py
+++ b/plugins/modules/mso_tenant.py
@@ -29,6 +29,10 @@ options:
   display_name:
     description:
     - The name of the tenant to be displayed in the web UI.
+    - On Nexus Dashboard (ND) 4.2+ / NDO 5.2+, O(display_name) must equal O(tenant); the API
+      rejects O(sites) association updates otherwise. When omitted on creation, it defaults to
+      O(tenant). Once set, the previous value is retained and resent on every update, so ensure
+      O(display_name) still matches O(tenant) before any update that changes O(sites).
     type: str
   description:
     description:
@@ -38,12 +42,18 @@ options:
     description:
     - A list of associated users for this tenant.
     - Using this property will replace any existing associated users.
-    - Admin user is always added to the associated user list irrespective of this parameter being used.
+    - Admin user is always added to the associated user list irrespective of this parameter being used, except on
+      Nexus Dashboard (ND) 4.2+ / NDO 5.2+ (see the deprecation note below).
+    - On ND 4.2+ / NDO 5.2+, user associations are derived from tenant-domain membership and the platform
+      automatically and immutably associates certain users (for example, superusers in the built-in
+      all-tenants-domain) with every tenant. This parameter is deprecated on ND 4.2+ / NDO 5.2+, is no longer
+      able to reliably manage user associations, and may be removed in a future version.
     type: list
     elements: str
   remote_users:
     description:
     - A list of associated remote users for this tenant.
+    - This parameter is deprecated on Nexus Dashboard (ND) 4.2+ / NDO 5.2+ for the same reasons as O(users).
     type: list
     elements: dict
     suboptions:
@@ -131,7 +141,7 @@ RETURN = r"""
 """
 
 from ansible.module_utils.basic import AnsibleModule
-from ansible_collections.cisco.mso.plugins.module_utils.mso import MSOModule, mso_argument_spec, ndo_remote_user_spec
+from ansible_collections.cisco.mso.plugins.module_utils.mso import MSOModule, mso_argument_spec, ndo_remote_user_spec, is_platform_version_at_least
 from ansible_collections.cisco.mso.plugins.module_utils.constants import YES_OR_NO_TO_BOOL_STRING_MAP
 
 
@@ -163,6 +173,7 @@ def main():
     orchestrator_only = module.params.get("orchestrator_only")
     state = module.params.get("state")
     remote_users = module.params.get("remote_users")
+    users = module.params.get("users")
 
     mso = MSOModule(module)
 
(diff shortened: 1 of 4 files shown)


plugins/modules/mso_tenant.py
```diff
@@ -216,7 +242,31 @@ def main():
...
         if mso.existing:
-            if mso.check_changed():
+            # ND returns extra cloud-account keys (awsAccount, azureAccount,
+            # gcpAccount, gatewayRouter) on each siteAssociations entry that this module never
+            # manages/sends (lookup_sites() only builds siteId/securityDomains). Left as-is, the
+            # exact-match list-of-dicts comparison in check_changed()/issubset() would report a
+            # perpetual false-positive "changed" for any tenant with a site association, so these
+            # nested keys are excluded from the comparison (the rest of each siteAssociations
+            # entry, e.g. siteId/securityDomains, is still compared).
+            ignore_keys = [
```
akinross: why use dotted notation if you split in https://github.com/CiscoDevNet/ansible-mso/pull/765/changes#diff-4ba7c8abb7103c0e8e3dfe658c6ec0a13e303a01cb61bce845ab61c31f83176eR1765? could we not pass in a list?

akinross (approved):

sajagana (approved):
LGTM!

